PR #12, opened 2018-02-21
Add Asset Types Models #155322116

wanjiku-carol:
**What does this PR do?**
Adds an Asset Type model to the models

**Description of work done?**
- Create an Asset Types Model
- Add tests for the model
- Register Model in the `admin.py`

> $ python manage.py test

[#155322116](https://www.pivotaltracker.com/story/show/155322116)


wanjiku-carol:
@HawiCaesar @jmnyarega Please review this PR

art/api/tests/tests_models.py
```diff
@@ -0,0 +1,31 @@
...
+    def setUp(self):
+        test_asset = AssetType(type="TestAsset")
+        test_asset.save()
+
+    def test_add_new_asset_type(self):
+        """Test add new asset type in model"""
+        new_assettype = AssetType(type="Macbook")
+        new_assettype.save()
+        all_assettypes = AssetType.objects.all()
+        self.assertEqual(len(all_assettypes), 2)
```
egichuri: we should also check before and after creation to confirm an increase in asset count. (you can also do `AssetType.objects.count()` instead of fetching all then getting the `len`)

art/api/tests/tests_models.py
```diff
@@ -0,0 +1,31 @@
...
+        get_asset.save()
+        self.assertEqual(len(AssetType.objects.all()), 1)
+        get_asset = AssetType.objects.get(type="New Name")
+        self.assertEqual(get_asset.type, "New Name")
+
+    def test_delete_asset_type(self):
+        """Test delete an asset type in model"""
+        get_asset = AssetType.objects.get(type="TestAsset")
+        get_asset.delete()
+        self.assertEqual(len(AssetType.objects.all()), 0)
```
egichuri: same here.. compare count before and after deletion.

art/api/models.py
```diff
@@ -1,3 +1,8 @@
 from django.db import models
 
-# Create your models here.
+class AssetType(models.Model):
+    """Stores all asset types"""
+    type = models.CharField(max_length=500)
```
egichuri: we might need to go with a different name since `type` is a built-in function in python..

egichuri: lets also have `created_at` and `last_modified` ([for ref](https://github.com/AndelaOSP/art-backend/blob/5024bb35614422762890243446f9bd334055a53e/art/api/models.py#L7))

art/art/settings.py
```diff
@@ -37,9 +37,7 @@
 SECRET_KEY = config('SECRET_KEY')
 DEBUG = config('DEBUG', default=False, cast=bool)
 DATABASES = {
-    'default': dj_database_url.config(
-        default=config('DATABASE_URL')
-    )
+    'default': dj_database_url.config()
```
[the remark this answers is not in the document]
egichuri: i'm ok with this since `dj_database_url ` checks for the `DATABASE_URL ` env variable by default..

art/api/models.py
```diff
@@ -1,3 +1,8 @@
 from django.db import models
 
-# Create your models here.
+class AssetType(models.Model):
```
jmnyarega: @HawiCaesar does  `pycode`  has a `two linespace rule` before a class? I think pep has.

art/api/models.py
```diff
@@ -1,3 +1,8 @@
...
+    """Stores all asset types"""
+    type = models.CharField(max_length=500)
+
+    def __str__(self):
+        return '{}'.format(self.type)
```
jmnyarega: Why not just return 

```   return self.type```

File: art/api/models.py (hunk 5 above)
Comment (HawiCaesar) [reply art/api/models.py]: @jmnyarega pycodestyle is PEP8. Check out this [link](https://pypi.python.org/pypi/pycodestyle/1.8.0.dev0). @wanjiku-carol it looks like you have not installed the pre-commit hook to check for PEP8 violations.

File: art/api/models.py (hunk 3 above)
Comment (HawiCaesar) [reply art/api/models.py]: +1 to having `created_at` and `last_modified`

**Comment by HawiCaesar** [comment]
As @egichuri said, let us move to having our PR titles like so `[#1234] enable xyz `

File: art/api/tests/tests_models.py (hunk 2 above)
Comment (wanjiku-carol) [reply art/api/tests/tests_models.py]: Will do

File: art/api/tests/tests_models.py (hunk 1 above)
Comment (wanjiku-carol) [reply art/api/tests/tests_models.py]: Will do

File: art/api/models.py (hunk 5 above)
Comment (wanjiku-carol) [reply art/api/models.py]: @HawiCaesar I believe there should be two spaces after imports.

File: art/api/models.py [hunk 7]
```diff
@@ -1,3 +1,11 @@
 from django.db import models
 
-# Create your models here.
+
+class AssetType(models.Model):
+    """Stores all asset types"""
+    asset_type = models.CharField(max_length=500)
+    created_at = models.DateTimeField(auto_now_add=True)
+    last_mofied = models.DateTimeField(auto_now_add=True)
```
egichuri: typo here.. should be `last_modified`

PR #13, opened 2018-02-22
[#155322114] Add Asset Category Model 

HawiCaesar:
### What does this PR do ?
-  Add the asset category model

`python manage.py test`

### Description of work done
- Create a asset category model
- Create tests for the asset category model

### Relelvant PT story
[#155322114](https://www.pivotaltracker.com/story/show/155322114)

### DIFF
--- a/art/api/models.py
+++ b/art/api/models.py
@@ -1,3 +1,11 @@
 from django.db import models
 
-# Create your models here.
+
+class AssetCategory(models.Model):
+    """ Stores all asset categories """
+    category_name = models.CharField(max_length=40, null=False)
+    created_at = models.DateTimeField(auto_now_add=True, editable=False)
+    last_modified = models.DateTimeField(auto_now=True, editable=False)
+
+    def __str__(self):
+        return self.category_name
--- a/art/api/tests/test_asset_category_model.py
+++ b/art/api/tests/test_asset_category_model.py
@@ -1,0 +1,31 @@
+from django.test import TestCase
+from ..models import AssetCategory
+
+
+class AssetCategoryModelTest(TestCase):
+    """ Tests for the Asset Category Model """
+    def setUp(self):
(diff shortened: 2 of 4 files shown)


art/api/models.py
```diff
@@ -1,3 +1,11 @@
 from django.db import models
 
-# Create your models here.
+
+class AssetCategory(models.Model):
+    """ Stores all asset categories """
+    name = models.CharField(max_length=40, null=False)
```
egichuri: Can we use something like `category_name`? `name` doesn't sound descriptive enough :)

art/api/models.py
```diff
@@ -1,3 +1,11 @@
...
+    created_at = models.DateTimeField(auto_now_add=True, editable=False)
+    modified_at = models.DateTimeField(auto_now=True, editable=False)
```
egichuri: `last_modified`

art/api/tests/test_asset_category_model.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class AssetCategoryModelTest(TestCase):
+    """ Tests for the Asset Category Model """
+    def setUp(self):
+        AssetCategory.objects.create(name="Accessories")
+        self.category = AssetCategory.objects.get(name="Accessories")
+
+    def test_can_save_a_category(self):
+        self.assertIn("Accessories", self.category.name)
```
egichuri: we might need to create a new category here and check for an increase in the category count.

art/api/models.py
HawiCaesar: I see your point. I will change this.

art/api/tests/test_asset_category_model.py
HawiCaesar: Alright makes sense! I have changed this.

art/api/models.py
HawiCaesar: I have changed this.

PR #18, opened 2018-02-26
Set Up Multiple Environment Settings #155269107

wanjiku-carol:
**What does this PR do?**
Creates multiple environment setup options

**Description of work done?**
- Create settings folder
- Separate files for different environments
- Environmental settings

> $ APP_ENV=(environment) python manage.py runserver

[#155269107](https://www.pivotaltracker.com/story/show/155269107)

### DIFF
--- a/art/settings/base.py
+++ b/art/settings/base.py
@@ -35,20 +35,11 @@ STATICFILES_STORAGE = 'whitenoise.django.GzipManifestStaticFilesStorage'
 SECRET_KEY = 'SECRET_KEY'
 
 SECRET_KEY = config('SECRET_KEY')
-DEBUG = config('DEBUG', default=False, cast=bool)
 DATABASES = {
     'default': dj_database_url.config(
         default=config('DATABASE_URL')
     )
 }
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
-
-ALLOWED_HOSTS = [
-    'art-backend.herokuapp.com',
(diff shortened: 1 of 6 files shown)


art/settings/base.py
```diff
@@ -37,15 +37,17 @@
 SECRET_KEY = config('SECRET_KEY')
 DEBUG = config('DEBUG', default=False, cast=bool)
```
egichuri: we can remove this debug since it's in the env specific settings

art/settings/base.py
```diff
@@ -37,15 +37,17 @@
...
 DATABASES = {
-    'default': dj_database_url.config()
+    'default': dj_database_url.config(
+        default=config('DATABASE_URL')
+    )
 }
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
```
egichuri: this too

art/settings/base.py
```diff
@@ -37,15 +37,17 @@
...
 
 ALLOWED_HOSTS = [
```
egichuri: this should be env specific - `art-backend.herokuapp.com` for prod and `'127.0.0.1',  '0.0.0.0',` for dev

wanjiku-carol: Should I have them removed from `base.py` when I place them in the other envs?

egichuri: yea.. the values in prod or dev envs should not be in base

PR #19, opened 2018-02-26
Create Item Model #155322123

wanjiku-carol:
**What does this PR do?**
Creates and Item model

**Description of work done?**
- Create `test_item_model.py` and add tests
- Add Item Model to `models.py`

> $ python manage.py test

[#155322123](https://www.pivotaltracker.com/story/show/155322123)


art/api/tests/test_item_model.py
```diff
@@ -0,0 +1,35 @@
+from django.test import TestCase
+
+from ..models import Item
+
+
+class AssetTypeModelTest(TestCase):
```
egichuri: i think this should be `ItemModelTest`..?

art/api/models.py
```diff
@@ -32,3 +32,13 @@ class AssetMake(models.Model):
 
     def __str__(self):
         return self.make_label
+
+
+class Item(models.Model):
+    """Stores all items"""
+    item_name = models.CharField(max_length=50, null=False)
+    created_at = models.DateTimeField(auto_now_add=True, editable=False)
+    last_modified = models.DateTimeField(auto_now=True, editable=False)
```
egichuri: some fields are missing here.. 
- an asset must have either an asset code or serial number
- all other fields required -> category, sub-category, type, make, model

api/models.py
```diff
@@ -16,7 +17,7 @@ def __str__(self):
 
 class AssetType(models.Model):
     """Stores all asset types"""
-    asset_type = models.CharField(max_length=500)
+    asset_type = models.CharField(max_length=500, null=False)
```
egichuri: the default for `null` is `False`, here we can change only the `max_length`

api/models.py
```diff
@@ -32,3 +33,24 @@ class AssetMake(models.Model):
...
+
+class Item(models.Model):
+    """Stores all items"""
+    item_code = models.CharField(max_length=50, null=False)
+    serial_number = models.CharField(max_length=50, null=False)
+    created_at = models.DateTimeField(auto_now_add=True, editable=False)
+    last_modified = models.DateTimeField(auto_now=True, editable=False)
+
+    def clean(self):
+        raise ValidationError(('Missing item details.'), code='required')
```
egichuri: the check can be done here instead of on `save()`

PR #25, opened 2018-03-01
[#155322120] Associate type with sub-category

HawiCaesar:
### What does this PR do ?
- create a one-to-many relationship between type and category
- refactor type_model tests

`python manage.py test`

### Relevant PT story
[#155322120](https://www.pivotaltracker.com/story/show/155322120)


api/tests/test_asset_type_model.py
```diff
@@ -1,26 +1,30 @@
...
-        test_asset = AssetType(asset_type="TestAsset")
-        test_asset.save()
+        self.new_sub_category = AssetSubCategory(
+            sub_category_name="Computer Assessories")
+        self.new_sub_category.save()
+        AssetType.objects.create(asset_type="Headset",
+                                 asset_sub_category=self.new_sub_category)
         self.all_assettypes = AssetType.objects.all()
 
     def test_add_new_asset_type(self):
```
egichuri: can we also check that a type isn't created if not associated with a sub-category?

HawiCaesar: I think it is but isn't that trying to test Django's code ? I tried testing this by adding a type without adding an existing subcategory and an error is thrown. This is okay but capturing the error and asserting it is proving difficult.

egichuri: which error are you getting?

api/models.py
```diff
@@ -14,11 +14,23 @@ def __str__(self):
...
+        return self.sub_category_name
+
+
 class AssetType(models.Model):
     """Stores all asset types"""
     asset_type = models.CharField(max_length=500)
     created_at = models.DateTimeField(auto_now_add=True, editable=False)
     last_modified = models.DateTimeField(auto_now=True, editable=False)
+    asset_sub_category = models.ForeignKey(AssetSubCategory,
+                                           on_delete=models.CASCADE, default=1)
```
egichuri: not sure we should be assigning a default sub-category

HawiCaesar: I did not want to assign a default but I got this prompt when I did not assign a default subcategory
screen shot 2018-03-01 at 12 06 54
I got this prompt because I moved `AssetSubCategory` before `AssetType`. If I did not do this then I could not add the foreign key

egichuri: that happens when making a change to model relationships, but the one-off default should be provided on the console for the migration. if done on the model it can cause undesired effects.

HawiCaesar: Alright makes sense, I will change this.

api/tests/test_asset_type_model.py
HawiCaesar: I managed to have a test for this actually. I implemented with the `self.assertRaises`

egichuri: awesome!

PR #29, opened 2018-03-02
[#155322121] Associate make and type

HawiCaesar:
### What does this PR do ?
- create a one-to-many relationship between make and type
- refactor make_model tests
- create migration file for relationship

`python manage.py test`

### Relevant PT story
[#155322121](https://www.pivotaltracker.com/story/show/155322121)


api/models.py
```diff
@@ -45,6 +45,7 @@ class AssetMake(models.Model):
     make_label = models.CharField(max_length=40, null=False)
     created_at = models.DateTimeField(auto_now_add=True, editable=False)
     last_modified_at = models.DateTimeField(auto_now=True, editable=False)
+    asset_type = models.ForeignKey(AssetType, on_delete=models.CASCADE)
```
egichuri: i'm thinking we should not allow someone to delete a type that is referenced somewhere to avoid accidental deletions. maybe present the user a message to delete all associated makes first?

HawiCaesar: Yes, makes sense. Would both help ?: A message indicating that all makes associated with that type will be delete and `on_delete=models.CASCADE` on the model ?

If we go with another option like `on_delete=models.SET_NULL` wont that have probably unnecessary records relating to nothing in the database ?

PR #30, opened 2018-03-05
Associate Model Number with Item #155649838

wanjiku-carol:
**What does this PR do?**
Adds an association between an item and the item model number

**Description of work done?**
- Add a model number foreign key in the Item model
- Write tests for the association

> $ python manage.py test

[#155649838](https://www.pivotaltracker.com/story/show/155649838)


wanjiku-carol:
@egichuri @HawiCaesar @jmnyarega  Please review

api/migrations/0009_item_model_number.py
```diff
@@ -0,0 +1,22 @@
...
+        ('api', '0008_auto_20180301_1625'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='item',
+            name='model_number',
+            field=models.ForeignKey(
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
```
HawiCaesar: Looks like you have not changed the migration file to be `models.PROTECT`

egichuri: i think that's catered for [here](https://github.com/AndelaOSP/art-backend/pull/30/files#diff-e6a82ebaf34ad7d12411085a5838ea38R17)

HawiCaesar: I see. What I have been doing is deleting the migration file, but this is fine.

api/migrations/0010_auto_20180305_1252.py
```diff
@@ -0,0 +1,21 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='item',
+            name='model_number',
+            field=models.ForeignKey(
+                null=True, on_delete=django.db.models.deletion.PROTECT,
+                to='api.ItemModelNumber'),
+        ),
```
wanjiku-carol: @HawiCaesar ☝️

HawiCaesar: True you did! When I was correcting the make model to include `models.PROTECT` I was deleting the migration file. So I have only one migration file instead of 2. But this works just as well.

PR #35, opened 2018-03-07
[#155322122] add missing migrations

egichuri:
https://www.pivotaltracker.com/story/show/155322122


api/migrations/0011_itemmodelnumber_make_label.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.0.1 on 2018-03-07 12:23
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0010_merge_20180306_1126'),
```
HawiCaesar: Won't this cause a migration file merge conflict with [this](https://github.com/AndelaOSP/art-backend/blob/develop/api/migrations/0010_merge_20180306_1126.py) on develop

egichuri: it defines the one on develop as a dependency.. the new one here (0011) depends on the one on develop (0011)

HawiCaesar: Ah, I didn't see the name of the file

PR #36, opened 2018-03-07
[#155512028] admin can add users

iankigen:
#### What does this PR do?
 - add a custom auth model
- enable admin to add users
- enable admin to see users on the admin dashboard

- create a custom auth model
- write test for the implementation

`python manage.py test`

All the migrations and tables that existed before creation of the custom auth model had to be deleted so as to remove traces of the `default auth` for the `custom auth` to work.

[#155512028](https://www.pivotaltracker.com/n/projects/2146417/stories/155512028)

#### Screenshots (if appropriate)
create_super_user
admin-login
add-user
all-users
change-user


iankigen:
@egichuri @HawiCaesar @jmnyarega @wanjiku-carol @nelmandela  Kindly review

api/models.py
```diff
@@ -86,3 +87,104 @@ def save(self, *args, **kwargs):
...
+            cohort=cohort,
+            slack_handle=slack_handle,
+            password=password,
+            is_staff=True,
+            is_admin=True
+        )
+        return user
+
+
+class User(AbstractBaseUser):
```
egichuri: any reason you chose to go with `AbstractBaseUser ` instead of `AbstractUser `?

iankigen: @egichuri  `AbstractUser` does not allow direct assignment of the `email` as a `USERNAME_FIELD`, It also has a default `username` field which we don't need as the `USERNAME_FIELD` in our context. 

`AbstractBaseUser` offers more flexibility by allowing creation of a custom `Manager` which may be helpful in the long run.

egichuri: if you use `AbstractUser ` the field you assign to `USERNAME_FIELD ` should override whatever is set on `AbstractUser`. Then you can set the `username` field to `None` on the custom user model. That way you don't need to have custom forms as well. Also for now the only changes i've seen on the custom `Manager` is removal of the username field.
something like 
```
class User(AbstractUser):
    username = None
    email = models.EmailField(_('email address'), unique=True)

    USERNAME_FIELD = 'email'
```
Lemme know if you wanna pair on this.

api/models.py
```diff
@@ -89,3 +90,104 @@ def save(self, *args, **kwargs):
     def __str__(self):
         return '{}{}{}'.format(self.item_code, self.serial_number,
                                self.model_number)
+
+
+class UserManager(BaseUserManager):
+    def create_user(self, email, name, cohort, slack_handle,
```
egichuri: these fields can be passed as kwargs

api/models.py
```diff
@@ -89,3 +90,104 @@ def save(self, *args, **kwargs):
...
+            cohort=cohort,
+            slack_handle=slack_handle,
+        )
+
+        user.set_password(password)
+        user.picture = picture
+        user.staff = is_staff
+        user.admin = is_admin
+        user.active = is_active
+        user.phone_number = phone_number
```
egichuri: we can also pass these fields to `self.model` since they're defined on the models level. and if we pass them as kwargs it'd make it easier. (see [here](https://github.com/django/django/blob/master/django/contrib/auth/models.py#L142))

api/models.py
```diff
@@ -89,3 +90,104 @@ def save(self, *args, **kwargs):
...
+        user.save(using=self._db)
+        return user
+
+    def create_staffuser(self, email, name, cohort,
+                         slack_handle, password=None):
+        user = self.create_user(
```
egichuri: We can use the default way it's done on django ([set_defaults](https://github.com/django/django/blob/master/django/contrib/auth/models.py#L154)) here, and remove `username`

api/models.py
```diff
@@ -89,3 +90,104 @@ def save(self, *args, **kwargs):
...
+            password=password,
+            is_staff=True,
+            is_admin=True
+        )
+        return user
+
+
+class User(AbstractBaseUser):
+    email = models.EmailField(max_length=50, unique=True)
+    name = models.CharField(max_length=50, blank=True, null=True)
```
egichuri: i'm thinking we go with first and last name since it's easier to work with that over the full name

api/models.py
iankigen: @egichuri I would love to pair with you on this.

api/models.py
iankigen: I agree

api/models.py
iankigen: Noted

api/tests/test_user.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class UserTestCase(TestCase):
+    def setUp(self):
+        self.user = User.objects.create(
+            email='test@site.com', name='test_user',
+            cohort=20, slack_handle='@test_user')
+
+    def test_can_add_user(self):
+        users_count_before = User.objects.all().count()
```
HawiCaesar: Any reason why you did not use `User.objects.count()` . It seem shorter and serves the same purpose. Same to line 22

api/models.py
```diff
@@ -89,3 +90,58 @@ def save(self, *args, **kwargs):
     def __str__(self):
         return '{}{}{}'.format(self.item_code, self.serial_number,
                                self.model_number)
+
+
+class UserManager(BaseUserManager):
+    use_in_migrations = True
+
+    def _create_user(self, **fields):
```
egichuri: let's also have tests for these methods. you can call `User.objects._create_user`, `User.objects.create_user`, `User.objects.create_superuser` and assert a user is created, and values in  `is_staff` and `is_superuser`

iankigen: I'm on it

iankigen:
@HawiCaesar @Awinja-Andela @nelmandela @jmnyarega Kindly review

egichuri:
approving this, but we should investigate the accuracy of the coverage.

PR #39, opened 2018-03-13
Coverage without nose

iankigen:
#### What does this PR do?
adds coverage without nose


iankigen:
@egichuri Kindly approve

PR #41, opened 2018-03-19
[#155922716] Admin can add security users

HawiCaesar:
### What does this PR do ?
- add security model
- add security model to the admin dashboard

- `python manage.py test`
- Login into your admin account
- Go to the `SecurityUser` section
- Add a user and click save


HawiCaesar:
@egichuri @alexkiura Is this a good approach to adding security personnel ? I am not too sure about what details should be in the security model.

alexkiura:
@HawiCaesar any reason we're not inheriting from the User model defined here? [here](https://github.com/AndelaOSP/art-backend/blob/develop/api/models.py#L135).

HawiCaesar:
@alexkiura I did consider it, though I had an error. I have managed to curb it. I am looking into having the admin UI hide a few unnecessary details for the user

api/tests/test_security_user_model.py
```diff
@@ -0,0 +1,46 @@
...
+    """ Tests for the Security User Model """
+    def setUp(self):
+        SecurityUser.objects.create(
+            email='security@site.com',
+            cohort=0,
+            slack_handle='@test_security_user',
+            password='devpassword',
+            first_name="TestFirst",
+            last_name="TestLast",
+            phone_number="254720900900",
```
egichuri: i don't think security users will have cohort, email address and slack handle.. i'm thinking we can use `AbstractUser`, but modify `UserManager` to have `_create_securitydesk_user` for security users and  `_create_user` for other users

HawiCaesar: Makes sense to modify `UserManager`. If we inherit from the User model, the email address is the username and it is required and I think it will almost always give an error.

egichuri: and if you specify the `USERNAME_FIELD ` for this model to be the phone number?

HawiCaesar: I still the error `This field is required.`

egichuri: Have you pushed your latest changes?

HawiCaesar: I have  now

HawiCaesar:
@alexkiura , I have managed to push the changes we had done together. Though when I try and add a new security user I get an error

```terminal
django.db.utils.IntegrityError: duplicate key value violates unique constraint "api_user_email_key"
DETAIL:  Key (email)=() already exists.
`
My tests fail as well cause of the above error

This is coming from the `User` model specifically the email field cause rule is that it should be `unique=True`. As mentioned by @egichuri the security team will not need an email address. But the User model cannot have more than one email address without an email. Should we disable it for now and get another unique identifier ?

File: api/models.py [hunk 2]
```diff
@@ -145,3 +148,30 @@ class User(AbstractUser):
...
+def set_username(sender, instance, *args, **kwargs):
+    if not instance.username:
+        instance.username = instance.email
+        instance.save()
+
+
+@receiver(post_save, sender=SecurityUser)
+def set_email(sender, instance, *args, **kwargs):
+    if not instance.email:
+        user_email = "{}@security.com".format(instance.badge_number)
```
egichuri: i think we should use a domain that's not in use.

HawiCaesar: So back to `example.com` or `andelasecurity.com` ?

egichuri: i'd go with example 

example

HawiCaesar: Now there is another thing I have learnt today

HawiCaesar: Let me change this

egichuri: you're welcome 😆

alexkiura:
@HawiCaesar Since SecurityUser inherits from User, we could override the email field attribute and disable the `unique=True` flag.

HawiCaesar:
@alexkiura I tried that from the start but got this error `django.core.exceptions.FieldError: Local field 'email' in class 'SecurityUser' clashes with field of the same name from base class 'User'`

Here are some links I saw 
[Stackoverflow](https://stackoverflow.com/a/2357942/4218361) and on the same [question](https://stackoverflow.com/a/39278537/4218361)
[Field Name Hiding Is Not Permitted](https://docs.djangoproject.com/en/2.0/topics/db/models/#field-name-hiding-is-not-permitted)

alexkiura:
@HawiCaesar available to sync on this?

api/models.py
```diff
@@ -147,3 +148,22 @@ class User(AbstractUser):
...
+
+    USERNAME_FIELD = 'badge_number'
+    REQUIRED_FIELDS = ['first_name', 'last_name',
+                       'badge_number']
+
+    class Meta:
+        verbose_name = "Security User"
+
+    def save(self, *args, **kwargs):
+
```
alexkiura: Let's get rid of the extra space.

api/admin.py
```diff
@@ -13,11 +17,45 @@
...
+    list_filter = (
+        'badge_number',
+    )
+
+    fieldsets = (
+        ('Account', {'fields': ('first_name', 'last_name',
+                                'badge_number',
+                                'phone_number',
+                                'password')}),
+
```
alexkiura: Let's remove this extra line as well.

api/tests/test_security_user_model.py
```diff
@@ -0,0 +1,55 @@
...
+            first_name="NewPerson",
+            last_name="LastName",
+            phone_number="254720400400",
+            badge_number="AE34"
+        )
+        new_security_team = SecurityUser.objects.count()
+
+        new_security_member.delete()
+        self.assertEqual(new_security_team, 2)
+        self.assertEqual(SecurityUser.objects.count(), 1)
```
alexkiura: And add a newline here.

HawiCaesar: Won't that be adding 2 new lines ? PEP8 flags adding a new line when there already is a new line as a violation.

api/models.py
```diff
@@ -147,3 +148,22 @@ class User(AbstractUser):
     USERNAME_FIELD = 'email'
     REQUIRED_FIELDS = ['cohort', 'slack_handle']
     objects = UserManager()
+
```
egichuri: while we're here.. can we add `verbose_name_plural = 'All Users'` for the User class? seems like it's also listing security users under `Users`

HawiCaesar: Sure. I will add this.

api/models.py

api/admin.py
HawiCaesar: Done

api/tests/test_security_user_model.py
alexkiura: My bad. It's hard to tell from the diff whether there's a new line. You can disregard this change.

PR #43, opened 2018-03-21
[#155922717] Add item status to item model

HawiCaesar:
### What does this PR do ?
- Add item status to item model

### Description of work done
- Add item status to item model
- Add fixed status file
- Add test for item status

### How can this be manually tested
- `python manage.py test`
- Login to the admin dashboard. 
- Click on Item. 
- Fill in the required details for the item. 
- The last field is the status. It includes `Available`, `Allocated` , `Lost`,  `Damaged`. Choose one of them
- Click Save


api/models.py
```diff
@@ -87,7 +101,11 @@ def save(self, *args, **kwargs):
         if not self.item_code and not self.serial_number:
             self.full_clean()
         else:
-            super(Item, self).save(*args, **kwargs)
+            if (self.status in status for status in status_list):
+                super(Item, self).save(*args, **kwargs)
+
+            else:
+                raise ValueError('Status provided does not exist')
```
egichuri: thinking we should raise this one in `clean` as well, and do `full_clean` by default.

HawiCaesar: I am not sure what you mean here...

HawiCaesar: Do you mean we can do `raise ValueError('Status provided does not exist')` in the clean and then after all the if-else add `self.full_clean` in the `clean` ?

api/models.py
```diff
@@ -1,6 +1,7 @@
 from django.db import models
 from django.core.exceptions import ValidationError
 from django.contrib.auth.models import AbstractUser, BaseUserManager
+from .statuses import status_list
```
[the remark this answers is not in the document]
HawiCaesar: I will change this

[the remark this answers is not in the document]

api/models.py
HawiCaesar: @egichuri Do you mean like what I have pushed ?

egichuri: uhmm.. not exactly..
in `clean`:
```
if (self.status not in status for status in status_list):
            raise ValueError('Status provided does not exist')
```
and in `save`:
```
self.full_clean()
        super(Item, self).save(*args, **kwargs)
```

egichuri: also, note `status_list` is a list already, so `self.status not in status_list` should work

api/models.py
```diff
@@ -68,30 +69,54 @@ def __str__(self):
...
     """Stores all items"""
+
+    status_list = ["Available", "Allocated", "Lost", "Damaged"]
+
+    item_statuses = (
+        (status_list[0], "Available"),
+        (status_list[1], "Allocated"),
+        (status_list[2], "Lost"),
+        (status_list[3], "Damaged")
+    )
```
egichuri: from [here](https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.Field.choices) -> "Generally, it’s best to define choices inside a model class, and to define a suitably-named constant for each value".. i think that'd be cleaner.

api/models.py
```diff
@@ -1,6 +1,7 @@
 from django.db import models
 from django.core.exceptions import ValidationError
 from django.contrib.auth.models import AbstractUser, BaseUserManager
+# from .statuses import status_list
```
egichuri: remember to delete this.

api/models.py
```diff
@@ -68,30 +68,53 @@ def __str__(self):
...
+    status = models.CharField(max_length=9,
+                              choices=item_statuses,
+                              default="Available")
 
     def clean(self):
         if not self.item_code and not self.serial_number:
             raise ValidationError(('Please provide either the serial number,\
                                asset code or both.'), code='required')
 
+        elif self.status in self.item_statuses:
```
egichuri: One last change: since `self.item_statuses` is a tuple of paired tuples you might need `dict(self.item_statuses)` for this to work.

PR #47, opened 2018-03-28
[#156224442] Add email address for security users

HawiCaesar:
### What does this PR do ?
- Allow the admin to add an email address for security users

### Description of work done
- change admin dashboard view to include email address
- remove code in `def save()`  that adds custom email address in security model
- refactor tests to include email address

### Screenshots if appropriate
screen shot 2018-03-28 at 11 54 11

### Relevant PT Story
[#156224442](https://www.pivotaltracker.com/story/show/156224442)

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -187,5 +187,4 @@ class SecurityUser(User):
         verbose_name = "Security User"
 
     def save(self, *args, **kwargs):
-
         super().save(*args, **kwargs)
--- a/api/models.py
+++ b/api/models.py
@@ -187,8 +187,5 @@ class SecurityUser(User):
         verbose_name = "Security User"
 
     def save(self, *args, **kwargs):
-        if not self.email:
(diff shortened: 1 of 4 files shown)


alexkiura:
@HawiCaesar that's not necessary. We can add that file to .coveragerc to prevent it from being analysed.

api/models.py
```diff
@@ -187,8 +187,4 @@ class Meta:
         verbose_name = "Security User"
 
     def save(self, *args, **kwargs):
```
egichuri: do we still need this?

HawiCaesar: I left it cause I guess you mentioned this `After discussions with the android team it seems like we'll not be implementing use of SMSs for now`

egichuri: yeah.. so i think we can get rid of the `save` methods since we don't have any more customizations.

HawiCaesar: Alright.

PR #48, opened 2018-04-03
[#156267228] User can see security users email addresses

HawiCaesar:
### What does this PR do ?
A user can see all security users (email addresses) via the API
### Description of work done
- add security serializer
- add security viewset
- add security url
- add test to list security users

- `python manage.py test`
- Via Postman: 
     - Login to the API using your credentials
     - go to the route `http://localhost:8000/api/v1/security_users/`
     -  You will see a list of security users depending on whether you have added a security user in your admin dashboard

### Screenshots 
screen shot 2018-04-05 at 17 40 02

screen shot 2018-04-03 at 15 39 46

### Relevant PT story
[#156267228](https://www.pivotaltracker.com/story/show/156267228)

### DIFF
--- a/api/tests/test_security_user.py
+++ b/api/tests/test_security_user.py
@@ -1,12 +1,14 @@
-from .test_user import UserTestCase, client, patch
+from django.test import TestCase
+from rest_framework.reverse import reverse
+from rest_framework.test import APIClient
+
 from ..models import SecurityUser
+client = APIClient()
 
 
-class SecurityUserTestCase(UserTestCase):
+class SecurityUserTestCase(TestCase):
     def setUp(self):
-        UserTestCase.setUp(self)
-
-        self.security_users = "/api/v1/security_users/"
+        self.security_users_url = reverse('security_users-list')
 
         SecurityUser.objects.create(
             email="sectest1@andela.com",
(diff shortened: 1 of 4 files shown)


HawiCaesar:
@iankigen @alexkiura Kindly review

api/tests/test_security_user.py
```diff
@@ -0,0 +1,40 @@
...
+            email="sectest2@andela.com",
+            password="devpassword",
+            first_name="TestFirst2",
+            last_name="TestLast2",
+            phone_number="254720900900",
+            badge_number="AE24"
+        )
+
+    def test_admin_can_get_security_users_emails(self):
+        client.login(username='admin@site.com', password='devpassword')
```
egichuri: @HawiCaesar as mentioned on the ticket this endpoint is for use within the android codebase, not when a user is logged in. it will use an API key which we'll save as an env variable on both ends. lemme know if this makes sense.

HawiCaesar: @egichuri it makes sense. But how will write a test for the endpoint? Or will the test work with the assumption of a fake key ?

egichuri: you should be able to test it once you set the authentication class(es) on the viewset.

HawiCaesar:
@iankigen @alexkiura Kindly review

api/views.py
```diff
@@ -31,3 +33,20 @@ def get_object(self):
         queryset = Item.objects.filter(assigned_to=self.request.user)
         obj = get_object_or_404(queryset, serial_number=self.kwargs['pk'])
         return obj
+
+
+class SecurityUserViewSet(ModelViewSet):
+    serializer_class = SecuritySerializer
+    permission_classes = [IsAuthenticated, ]
+    authentication_classes = (FirebaseTokenAuthentication,)
```
egichuri: as per our discussion, let's have a different authentication class here.

HawiCaesar: Okay, then should this ticket still include writing how the authentication should work? I think that should be a ticket on its own. What do you think ?

api/views.py
```diff
@@ -31,3 +33,18 @@ def get_object(self):
...
         return obj
+
+
+class SecurityUserViewSet(ModelViewSet):
+    serializer_class = SecuritySerializer
+    http_method_names = ['get']
+
+    def list(self, request, *args, **kwargs):
+        security_emails = SecurityUser.objects.all()
+
```
alexkiura: @HawiCaesar consider using a list comprehension here . It's much faster and helps avoid making two lists and doing the append. This could be refactored as:
```
list_of_emails = [security_user.email for security_user in SecurityUser.objects.all()]
```

PR #50, opened 2018-04-10
feat(checkin-checkout): security team member can log asset checkin ch…

nelmandela:
### What does this PR do?

  •Security team member can log asset checkin/checkout  

### Description of work done

  •Add checkin model
  •Add checkout model
  •Add tests 

*python manage.py test*
  •Via Postman:
      *Login to the API using your credentials*
      *go to the route http://localhost:8000/api/v1/checkin/ or  http://localhost:8000/api/v1/checkout/* 
      *Post checkin/checkout data or get checkin/checkout data* 
Relevant PT story
#155922721


nelmandela:
@egichuri thanks for that.we are on it

nelmandela:
@egichuri done

api/admin.py
```diff
@@ -6,7 +6,9 @@
                      Item,
                      AssetMake,
                      ItemModelNumber,
-                     SecurityUser)
+                     SecurityUser,
+                     CheckinCheckout
```
egichuri: lets use `AssetLog` instead of `CheckinCheckout`

api/authentication.py
```diff
@@ -33,3 +33,16 @@ def authenticate_credentials(self, key):
         if not user.is_active:
             raise exceptions.AuthenticationFailed('User inactive or deleted.')
         return (user, token)
+
+
+class IsSecurityUser(BasePermission):
```
egichuri: this is better suited in `api/permissions.py`

api/models.py
```diff
@@ -185,3 +185,29 @@ class SecurityUser(User):
...
+    CHECKOUT = "Checkout"
+
+    option = (
+        (CHECKIN, "Checkin"),
+        (CHECKOUT, "Checkout"),
+    )
+    item = models.ForeignKey(Item,
+                             null=False,
+                             on_delete=models.PROTECT)
+    security_user = models.ForeignKey(SecurityUser,
```
egichuri: can we name this field `logged_by` or `checked_by`?

api/models.py
```diff
@@ -185,3 +185,29 @@ class SecurityUser(User):
...
+                                      blank=True,
+                                      default=1,
+                                      on_delete=models.PROTECT)
+    option = models.CharField(max_length=10,
```
egichuri: and `log_type` for this

api/admin.py
iankigen: On it

api/authentication.py
iankigen: working on it

api/models.py
iankigen: I agree, working on it

api/authentication.py
```diff
@@ -33,3 +33,16 @@ def authenticate_credentials(self, key):
...
+    """
+    Allows access only to security users.
+    """
+
+    def has_permission(self, request, view):
+        try:
+            user = request.user.securityuser
+        except Exception:
+            return False
+        return request.user and user and request.user.is_authenticated
```
egichuri: `user and user.is_authenticated` should work here

api/models.py
```diff
@@ -185,3 +185,29 @@ class SecurityUser(User):
...
+    option = (
+        (CHECKIN, "Checkin"),
+        (CHECKOUT, "Checkout"),
+    )
+    item = models.ForeignKey(Item,
+                             null=False,
+                             on_delete=models.PROTECT)
+    checked_by = models.ForeignKey(SecurityUser,
+                                   blank=True,
+                                   default=1,
```
egichuri: we're not sure there'll always be a security user with id 1, so we shouldn't set a default here. 
also can we have this field take in the currently logged in user by default?

api/models.py
```diff
@@ -185,3 +185,29 @@ class SecurityUser(User):
...
+                                   on_delete=models.PROTECT)
+    log_type = models.CharField(max_length=10,
```
egichuri: this should be required, no default

api/models.py
```diff
@@ -185,3 +185,29 @@ class SecurityUser(User):
...
+class AssetLog(models.Model):
+    """Stores checkin/Checkout asset logs"""
+    CHECKIN = "Checkin"
+    CHECKOUT = "Checkout"
+
+    option = (
+        (CHECKIN, "Checkin"),
+        (CHECKOUT, "Checkout"),
+    )
+    item = models.ForeignKey(Item,
```
egichuri: (sorry i should have added this to the initial request on the ticket but) can we make this reference the `serial_number` and not the item id? ( see [here](https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.ForeignKey.to_field))
note: for that to happen https://www.pivotaltracker.com/story/show/156196172 should be completed on this PR

api/urls.py
```diff
@@ -1,9 +1,12 @@
...
 
-from .views import UserViewSet, ItemViewSet, SecurityUserViewSet
+from .views import UserViewSet, ItemViewSet, SecurityUserViewSet, \
+    AssetLogViewSet
 
 router = SimpleRouter()
 router.register('users', UserViewSet)
 router.register('items', ItemViewSet, 'items')
 router.register('security_users', SecurityUserViewSet, 'security_users')
+router.register('asset-log', AssetLogViewSet, 'asset-log')
```
egichuri: `'asset-logs'`

api/models.py
iankigen: on it

api/models.py
iankigen: alright. will do that

alexkiura: Since going forward we want to use `asset` instead of `item` (see #156704243). Consider changing this to `asset = models.ForeignKey(Item...)`

nelmandela: @alexkiura since there's a high priority PT task for refactoring all item references to asset, can I continue using items for now since tests won't pass

If this task can be completed early I pick the refactoring task immediately thereafter

api/views.py
```diff
@@ -54,3 +54,8 @@ class AssetLogViewSet(ModelViewSet):
     permission_classes = [IsSecurityUser]
     authentication_classes = (FirebaseTokenAuthentication,)
     http_method_names = ['get', 'post']
+
+    def perform_create(self, serializer):
+        sec_user = SecurityUser.objects.get(id=self.request.user.id)
+        if self.request.user:
```
egichuri: you could do `sec_user = self.request.user.securityuser`, then remove the conditional since a user has to be present to access this serializer.

nelmandela:
@egichuri done

api/models.py
egichuri: @nelmandela this isn't done yet. "reference the item serial_number and not the item id"

egichuri: we should have `to_field='serial_number',` here

nelmandela: thanks for that. on it

iankigen:
@egichuri @alexkiura Kindly review

api/models.py
```diff
@@ -82,8 +82,8 @@ class Item(models.Model):
         (DAMAGED, "Damaged")
     )
 
-    item_code = models.CharField(max_length=50, blank=True)
-    serial_number = models.CharField(max_length=50, blank=True)
+    item_code = models.CharField(unique=True, max_length=50, blank=True)
```
alexkiura: Having both `unique=True` and `blank=True` as options for the same field will cause an `IntegrityError: duplicate key value violates unique constraint` error if `item_code` and `serial_number` is not provided. Consider doing away with `blank=True` if we're making these fields unique.
This applies for both item_code & serial_number.

api/permissions.py
```diff
@@ -5,10 +5,22 @@ class IsApiUser(BasePermission):
...
             app = request.auth.application
         except Exception:
             return False
         return app and not request.auth.user
+
+
+class IsSecurityUser(BasePermission):
+    """
+    Allows access only to security users.
+
```
alexkiura: Unnecessary newline here.

nelmandela:
@alexkiura

alexkiura:
@nelmandela looks good, but since we changed the model fields, remember to run a `make_migrations` to update the migrations.

iankigen:
@alexkiura Already did that

PR #55, opened 2018-04-16
[#156725831] User Feedback Endpoint

HawiCaesar:
### What does this PR do ?
User can submit feedback via API

### Description of work done
- add tests for user feedback model
- add tests for user feedback endpoint
- add user feedback model
- add user feedback viewset
- add user feedback serializer

- `python manage.py test`

### Relevant PT Story
[#156725831](https://www.pivotaltracker.com/story/show/156725831)


api/tests/test_user_feedback_api.py
```diff
@@ -0,0 +1,89 @@
...
+                                    report_type="bug")
+
+        self.feedback_url = '/api/v1/user-feedback/'
+        self.token_user = 'testtoken'
+
+    @patch('api.authentication.auth.verify_id_token')
+    def test_can_post_feedback(self, mock_verify_token):
+        mock_verify_token.return_value = {'email': self.user.email}
+        data = {
+            "reported_by": "test4@site.com",
```
egichuri: we shouldn't be having `reported_by` sent to the API endpoint.

HawiCaesar: It will come from the token ?

egichuri: i believe you're setting it [here](https://github.com/AndelaOSP/art-backend/pull/55/files#diff-265524153427c5c3f0dd9ae087f16a4bR72) :)

api/models.py
```diff
@@ -233,3 +233,20 @@ def clean(self):
...
+
+class UserFeedback(models.Model):
+    """ Stores user feedback data """
+    FEEDBACK = "feedback"
+    BUG = "bug"
+    option = (
+        (FEEDBACK, "feedback"),
+        (BUG, "bug"),
+    )
+    reported_by = models.EmailField(max_length=50, unique=False, null=False)
```
egichuri: the reporter should be an existing user

api/tests/test_user_feedback_api.py
```diff
@@ -0,0 +1,89 @@
...
+        self.user = User.objects.create(
+            email='test4@site.com', cohort=20,
+            slack_handle='@test_user4', password='devpassword'
+        )
+
+        UserFeedback.objects.create(reported_by="X.Y@andela.com",
+                                    message="This is feedback",
+                                    report_type="bug")
+
+        self.feedback_url = '/api/v1/user-feedback/'
```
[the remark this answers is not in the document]
HawiCaesar: Done.

api/models.py
HawiCaesar: But won't this always save an existing user as the issue of the token comes to play ?

I did try to modify the models to include a the `User` model but I get an extra column in the db and it makes its own user. With this approach I can only post once to the endpoint See the picture from my db
screen shot 2018-04-17 at 14 43 15
screen shot 2018-04-17 at 14 43 37

egichuri: i think it's ok to have the current user set as the one who's submitted the feedback since to access the app one has to be logged in. Not sure what you mean by the extra column..

egichuri: and by existing user i meant we should go with `ForeignKey` instead of `EmailField `

api/tests/test_user_feedback_api.py
```diff
@@ -0,0 +1,69 @@
...
+from rest_framework.test import APIClient
+from rest_framework.reverse import reverse
+
+
+from ..models import User
+
+client = APIClient()
+
+
+class UserFeedbackModelTest(TestCase):
```
egichuri: i think this should be API test

HawiCaesar: Right you are. Thanks for this!
Done.

api/models.py
```diff
@@ -233,3 +233,20 @@ def clean(self):
...
+class UserFeedback(models.Model):
+    """ Stores user feedback data """
+    FEEDBACK = "feedback"
+    BUG = "bug"
+    option = (
+        (FEEDBACK, "feedback"),
+        (BUG, "bug"),
+    )
+    reported_by = models.ForeignKey(User, on_delete=models.PROTECT)
+    message = models.CharField(max_length=200, unique=False, null=False)
```
egichuri: isn't unique False by default?

HawiCaesar: It actually is. Thanks for this. Let me change this.

PR #56, opened 2018-04-17
[#156725831] User can submit feedback via API

HawiCaesar:
### What does this PR do ?
Add get_queryset for user feedback endpoint

### Relevant PT Story
[#156725831](https://www.pivotaltracker.com/story/show/156725831)

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -73,12 +73,10 @@ class AssetLogViewSet(ModelViewSet):
 
 class UserFeedbackViewSet(ModelViewSet):
     serializer_class = UserFeedbackSerializer
+    queryset = UserFeedback.objects.all()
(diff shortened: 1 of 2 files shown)


api/views.py
```diff
@@ -79,3 +79,6 @@ class UserFeedbackViewSet(ModelViewSet):
 
     def perform_create(self, serializer):
         serializer.save(reported_by=self.request.user)
+
+    def get_queryset(self):
+        return UserFeedback.objects.all()
```
egichuri: since we're not modifying the queryset we can  have it after `    serializer_class = UserFeedbackSerializer `

HawiCaesar: Done.

PR #57, opened 2018-04-18
[#156704243] Rename all occurrences`item` to `asset`

HawiCaesar:
### What does this PR do ?
Rename item to asset 

### Description of work done
- rename all occurrences of Item (exact word and case match) to Asset (this is for the model)
- rename all ocurrences of ItemModelNumber (exact word and case match) to AssetModelNumber
- rename all occurences of item_code (exact word and case match) to asset_code
- rename all other occurences of item or Item to asset or Asset respectively eg test_itemmodel, item_statuses, etc
- run makemigrations

### Relevant PT Story
[#156704243](https://www.pivotaltracker.com/story/show/156704243)


egichuri:
@HawiCaesar did you run `makemigrations` after each step? we should be having 3 migrations files, otherwise we'll have a problem for existing records

HawiCaesar:
@egichuri We have sorted the problem out.

api/models.py
```diff
@@ -67,41 +67,41 @@ def __str__(self):
         return self.model_number
 
 
-class Item(models.Model):
+class Asset(models.Model):
     """Stores all items"""
```
egichuri: rename this as well

api/models.py
```diff
@@ -110,10 +110,10 @@ def save(self, *args, **kwargs):
         are provided and an existing status is given
```
egichuri: and this - line [109](https://github.com/AndelaOSP/art-backend/pull/57/files#diff-9dba586f38463d27c2d7fdadad3eb040R109)

api/models.py

api/models.py

PR #58, opened 2018-04-18
[#156877577] Make the password field optional

iankigen:
#### What does this PR do?

- Make the password field optional 

- add user creation form

`python manage.py test`

[ #156877577](https://www.pivotaltracker.com/n/projects/2146417/stories/156877577)

#### Screenshots
no_pass
no_pass_admin


egichuri:
@iankigen i think the password field is required by default. so we might need to override that on the `User` model as well

iankigen:
@egichuri by removing `password` from `UserSerializer` in my previous task, we could create a User without a password through the APi.

egichuri:
@iankigen yea.. but we need the db to allow that as well for the sake of users created outside of the API and admin form

HawiCaesar:
@egichuri in what case will a user be created outside the API and admin form ?

egichuri:
eg fixtures

api/models.py
```diff
@@ -168,6 +168,7 @@ class User(AbstractUser):
     picture = models.CharField(max_length=255, blank=True, null=True)
     phone_number = models.CharField(max_length=50, blank=True, null=True)
     last_modified = models.DateTimeField(auto_now=True, editable=False)
+    password = models.CharField(max_length=128, blank=True, default='')
```
egichuri: we could do `null=True` and get rid of the default

iankigen: on it

iankigen:
@egichuri done

PR #59, opened 2018-04-19
[156915963] Allow assets endpoint for all authenticated users

egichuri:
https://www.pivotaltracker.com/story/show/156915963


iankigen:
LGTM

PR #60, opened 2018-04-23
[#156990530] Make password field optional for security user

HawiCaesar:
### What does this PR do ?
Makes the password field optional for security users
### Description of work done
- write test for optional password
- add `SecurityUserRegistrationForm` class
- import `SecurityUserRegistrationForm`  and use it  in `SecurityUserAdmin` class
- `python manage.py runserver` AND , OR
- Login into the admin dashboard
- Add a security user and leave out the `password` and `password confirmation` fields.
- Click `Save`. You should see a `success` message indicating the user has been added.

### DIFF
--- a/api/admin.py
+++ b/api/admin.py
@@ -1,7 +1,7 @@
 from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
-from .forms import UserRegistrationForm, SecurityUserRegistrationForm
+from .forms import UserRegistrationForm
 from .models import (AssetCategory, AssetType,
                      AssetSubCategory,
                      Asset,
@@ -26,7 +26,7 @@ admin.site.register(
 
 
 class SecurityUserAdmin(BaseUserAdmin):
-    add_form = SecurityUserRegistrationForm
+    add_form = UserRegistrationForm
     list_display = (
         'first_name',
         'last_name',
--- a/api/admin.py
+++ b/api/admin.py
@@ -1,7 +1,7 @@
 from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
-from .forms import UserRegistrationForm
+from .forms import UserRegistrationForm, SecurityUserRegistrationForm
 from .models import (AssetCategory, AssetType,
                      AssetSubCategory,
                      Asset,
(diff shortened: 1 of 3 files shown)


api/forms.py
```diff
@@ -19,3 +19,7 @@ class UserRegistrationForm(UserCreationForm):
         help_text=_("Enter the same password as before, for verification."),
         required=False
     )
+
+
+class SecurityUserRegistrationForm(UserRegistrationForm):
```
alexkiura: Not sure the `SecurityUserRegistrationForm` class is necessary since it inherits from `UserRegistrationForm` and does not extend it in any way. 
We have two options here:
1. Set `add_form= UserRegistrationForm` in the `SecurityUserAdmin` ([here](https://github.com/AndelaOSP/art-backend/pull/60/files#diff-ee05abd68c2faa98e3f8f2ac878d3231R30)) and do away with the empty `SecurityUserRegistrationForm`
2. Create a `BaseUserRegistrationForm` that will be inherited by `UserRegistrationForm` and `SecurityUserRegistrationForm`. Each will customize it differently. This will definitely warrant more effort though.

However, since `UserRegistrationForm` is already defined and does what we want, I'd recommend option 1. #

api/admin.py
```diff
@@ -27,6 +27,7 @@
 
 
 class SecurityUserAdmin(BaseUserAdmin):
+    add_form = SecurityUserRegistrationForm
```
alexkiura: See my comments in the `SecurityUserRegistrationForm`.

api/forms.py
HawiCaesar: Alright going with option 1

api/tests/test_security_user_model.py
```diff
@@ -55,3 +55,19 @@ def test_can_delete_a_security_user(self):
...
+
+    def test_can_add_security_user_without_password(self):
+        SecurityUser.objects.create(
+            email="sectest98@andela.com",
+            first_name="TestNew",
+            last_name="TestLastNew",
+            phone_number="254720900901",
+            badge_number="AXW23"
+        )
+
```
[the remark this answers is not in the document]
HawiCaesar: Makes sense. I will change this.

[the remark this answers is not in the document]
HawiCaesar: Done.

api/forms.py
HawiCaesar: Done.

PR #66, opened 2018-04-27
[#157052270], [#156990275] User can track asset assignment, More details on Asset list page

HawiCaesar:
### What does this PR do ?
- Allows a system user to track asset assignment
- Allow the Admin to see more details on the Asset list page 
### Description of work done
- Create `AllocationsHistory` model 
- Add `asset`, `previous_owner`, `current_owner`,  `created_at` fields to model
- Add `post_save` method that makes a `current_owner` the `previous_owner` when reassignment is done
- write tests
- add `list_display` in `AssetAdmin` to show more details of an asset

### Screenshot
screen shot 2018-04-27 at 10 23 21

### Relevant PT story
[#157052270](https://www.pivotaltracker.com/story/show/157052270)
[#156990275](https://www.pivotaltracker.com/story/show/156990275)


api/migrations/0030_auto_20180426_1313.py
```diff
@@ -0,0 +1,48 @@
...
+            ],
+        ),
+        migrations.RemoveField(
+            model_name='asset',
+            name='assigned_to',
+        ),
+        migrations.AddField(
+            model_name='asset',
+            name='current_owner',
+            field=models.CharField(default=django.utils.timezone.now, editable=False, max_length=50),
```
egichuri: please change this default.. date isn't valid for an integer field

api/models.py
```diff
@@ -308,3 +345,10 @@ def save_initial_asset_status(sender, **kwargs):
         AssetStatus.objects.create(asset=current_asset,
                                    current_status="Available")
         current_asset.save()
+
+
+@receiver(post_save, sender=AllocationHistory)
+def save_current_asset_owner(sender, **kwargs):
+    asset_owner = kwargs.get('instance')
+    asset_owner.asset.current_owner = asset_owner.current_owner
```
egichuri: `asset_owner.asset.assigned_to`

api/models.py
```diff
@@ -292,6 +293,42 @@ def save(self, *args, **kwargs):
...
+    previous_owner = models.ForeignKey('User',
+                                       related_name='previous_owner_asset',
+                                       editable=False,
+                                       blank=True,
+                                       null=True,
+                                       on_delete=models.PROTECT)
+    created_at = models.DateTimeField(auto_now_add=True, editable=False)
+
+    class Meta:
+        verbose_name_plural = "Allocation Histories"
```
egichuri: "Allocation History" works better here

api/migrations/0030_auto_20180426_1313.py
HawiCaesar: Done.

api/models.py
HawiCaesar: I guess so but`histories` doesn't sound so bad does it ?
screen shot 2018-04-30 at 12 57 02

api/models.py
HawiCaesar: We have changed this.

api/models.py
egichuri: it's like app changelog vs changelogs.. :)

PR #69, opened 2018-05-02
[#157103974] User can track item status via API

iankigen:
#### What does this PR do?

Allows user to track item status via API

- add asset status serializer
- add asset status viewset
- add test

`python manage.py test`
`/asset-status/`

[#157103974](https://www.pivotaltracker.com/n/projects/2146417/stories/157103974)

#### Screenshots

post-asset

### DIFF
--- a/api/tests/test_asset_status_api.py
+++ b/api/tests/test_asset_status_api.py
@@ -1,0 +1,166 @@
+from unittest.mock import patch
+from django.test import TestCase
+from django.contrib.auth import get_user_model
+from rest_framework.reverse import reverse
+from rest_framework.test import APIClient
+
+from core.models import Asset, AssetModelNumber, AssetStatus
+
+User = get_user_model()
+client = APIClient()
+
+
+class AssetStatusAPITest(TestCase):
+    """Tests for the Asset Status API"""
+
+    def setUp(self):
+        self.test_assetmodel1 = AssetModelNumber(model_number="IMN50987")
+        self.test_assetmodel1.save()
+        self.token_user = 'testtoken'
+
+        self.normal_user = User.objects.create(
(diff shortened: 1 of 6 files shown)


api/tests/test_asset_status_api.py
```diff
@@ -0,0 +1,166 @@
...
+        }
+        response = client.post(
+            self.asset_status_urls,
+            data=data,
+            HTTP_AUTHORIZATION="Token {}".format(self.token_user))
+        self.assertEqual(response.data, {
+            'asset': ['Object with serial_number=Invalid does not exist.']
+        })
+
+        self.assertEqual(response.status_code, 400)
```
HawiCaesar: This is okay but how come the status code is not 404 ?

iankigen: Its supposed to be 400 not 404 because the server cannot process the request due to something that is perceived to be a client error which is true.

api/views.py
```diff
@@ -80,3 +81,11 @@ class UserFeedbackViewSet(ModelViewSet):
 
     def perform_create(self, serializer):
         serializer.save(reported_by=self.request.user)
+
+
+class AssetStatusViewSet(ModelViewSet):
+    serializer_class = AssetStatusSerializer
+    queryset = AssetStatus.objects.all()
+    permission_classes = [IsAuthenticated]
+    authentication_classes = (FirebaseTokenAuthentication,)
```
alexkiura: For consistency, let's have `authentication_classes = [FirebaseTokenAuthentication,] ` since we have already used a list to describe the `permission_classes`. Also, add a comma inside the permission_classes, so it looks like `permission_classes = [IsAuthenticated,]`

PR #70, opened 2018-05-02
[#156990322] Change asset log display in admin dashboard

HawiCaesar:
### What does this PR do ?
Change the Asset logs on admin dashboard to make sense

### Description of work done
- add `AssetLogsAdmin` class
- add `list_display` with date, asset, checked_by, log_type columns
- add `AssetLogsAdmin` to `admin.site.register`

- Login to the admin dashboard
- click on `Asset Logs`
- you will be able to see the Asset Logs table with its columns

### Screenshots
screen shot 2018-05-02 at 14 00 22

### Relevant PT Story
[#156990322](https://www.pivotaltracker.com/story/show/156990322)

### DIFF
--- a/core/admin.py
+++ b/core/admin.py
@@ -2,15 +2,16 @@ from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
 from .forms import UserRegistrationForm
-from .models.asset import (
-        AssetCategory, AssetType,
-        AssetSubCategory,
-        Asset,
-        AssetMake,
-        AssetLog,
-        AssetStatus,
-        AssetModelNumber,
-        AllocationHistory)
+from .models.asset import (AssetCategory,
+                           AssetType,
+                           AssetSubCategory,
+                           Asset,
+                           AssetMake,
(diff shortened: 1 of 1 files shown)


core/admin.py
```diff
@@ -125,9 +125,14 @@ class AllocationHistoryAdmin(admin.ModelAdmin):
     list_display = ('asset', 'current_owner', 'previous_owner', 'created_at')
 
 
+class AssetLogsAdmin(admin.ModelAdmin):
+    list_display = ('created_at', 'asset', 'checked_by', 'log_type')
```
alexkiura: @HawiCaesar I'm thinking for the date, we should pick `last_modified` instead of `created_at`. Since that would give us the most recent date a log was saved. What are your thoughts on this @egichuri ?

HawiCaesar: True @alexkiura, but in this case each asset log is its own record. Individual asset logs are not modified (`editable`) as per [here](https://github.com/AndelaOSP/art-backend/blob/develop/core/models/asset.py#L150). Both `created_at` and `last_modified` will contain the same data.

alexkiura: True. True.

PR #71, opened 2018-05-03
[#157183400] Change assigned_to to None when asset status is changed to available [WIP]

nelmandela:
### What does this PR do ?
After marking an asset as available, the current owner of the asset is changed to None

### Description of work done
- fix failing test for test_asset_api
- add if condition for post_save function
- add test in test_asset_status_model
### How do I manually test this ?
- `python manage.py test` OR
- Login into the dashboard. 
- assign an asset to a user in the `AllocationHistory`
- assign an asset to "Allocated" Status
- then assign the asset back to "Available"
- Go to the assets and the `assigned_to` column of that particular asset is marked with `-` meaning it is changed to `None`

### Screenshots
screen shot 2018-05-03 at 15 54 37
screen shot 2018-05-03 at 15 54 51
screen shot 2018-05-03 at 15 55 02

### Relevant PT Story
[#157183400](https://www.pivotaltracker.com/story/show/157183400)


core/models/asset.py
```diff
@@ -229,6 +229,8 @@ def save(self, *args, **kwargs):
 def set_current_asset_status(sender, **kwargs):
     asset_status = kwargs.get('instance')
     asset_status.asset.current_status = asset_status.current_status
+    if asset_status.current_status == "Available":
+        asset_status.asset.assigned_to = None
     asset_status.asset.save()
```
wanjiku-carol: Shouldn't this also set the former `current_owner` to be the new `previous_owner`?

egichuri: I think @wanjiku-carol's comment makes sense

PR #79, opened 2018-05-11
#157183398 Added asset Condition model  and tests

Awinja-Andela:
#### What does this PR do?
This pull request introduces Asset condition model and tests
This should be possible:
When adding an asset for the first time
When assigning an asset to a user
When the asset is returned

python manage.py test
User could not be able to write notes on the condition of the asset.
#157183398
#### Screenshots (if appropriate)
screen shot 2018-05-11 at 11 52 43 am


core/models/asset.py
```diff
@@ -112,9 +116,13 @@ class Asset(models.Model):
                                     editable=False,
                                     null=True,
                                     on_delete=models.PROTECT)
-    model_number = models.ForeignKey(AssetModelNumber, null=True,
+    model_number = models.ForeignKey(AssetModelNumber,
+                                     null=True,
                                      on_delete=models.PROTECT)
     current_status = models.CharField(editable=False, max_length=50)
+    condition_description = models.CharField(editable=False,
```
egichuri: can we call this `asset_condition`?

core/models/asset.py
```diff
@@ -225,6 +233,25 @@ def save(self, *args, **kwargs):
         super(AllocationHistory, self).save(*args, **kwargs)
 
 
+class AssetCondition(models.Model):
+    asset = models.ForeignKey(Asset,
+                              to_field="serial_number",
+                              null=False,
+                              on_delete=models.PROTECT)
+
+    condition_description = models.CharField(max_length=50,
```
egichuri: same here

core/models/asset.py
```diff
@@ -25,6 +25,10 @@
     (CHECKIN, "Checkin"),
     (CHECKOUT, "Checkout"),
 )
+NEW = "Brand New"
+WORKING = "Working"
+ISSUES = "Issues"
+NOT_WORKING = "Not Working"
```
egichuri: do we need these?

Awinja-Andela: Nope. Removed.

PR #80, opened 2018-05-14
[#157396331] User can see create and view asset model numbers via API

iankigen:
#### What does this PR do?

Allows user to create and view asset model number via API

#### Description of Task to be completed?

- add Asset Model Number Viewset
- add Asset Model Number Serializer
- add test

#### How should this be manually tested?

`python manage.py test`
`/asset-models/`

#### What are the relevant pivotal tracker stories?

[ #157396331](https://www.pivotaltracker.com/n/projects/2146417/stories/157396331)

### Screenshots

get
desktop

PR #81, opened 2018-05-14
[#157396330] User can see create and view asset makes via API 

andela-Taiwo:
#### What does this PR do?
- Allows the user to create and view asset make via API

- add test
- add AssetMake Serializer
- add AssetTMake Viewset

How should this be manually tested?
`python manage.py test`

What are the relevant pivotal tracker stories?
#157396330

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -7,13 +7,13 @@ from rest_framework.viewsets import ModelViewSet
 from api.authentication import FirebaseTokenAuthentication
 from core.models import Asset, SecurityUser, AssetLog, UserFeedback, \
     AssetStatus, AllocationHistory, AssetCategory, AssetSubCategory, \
-    AssetType, AssetModelNumber, AssetCondition
+    AssetType, AssetModelNumber, AssetCondition, AssetMake
 from .serializers import UserSerializer, \
     AssetSerializer, SecurityUserEmailsSerializer, \
     AssetLogSerializer, UserFeedbackSerializer, \
     AssetStatusSerializer, AllocationsSerializer, AssetCategorySerializer, \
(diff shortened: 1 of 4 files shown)


api/views.py
```diff
@@ -120,3 +120,11 @@ class AssetTypeViewSet(ModelViewSet):
...
     authentication_classes = (FirebaseTokenAuthentication,)
     http_method_names = ['get', 'post']
+
+
+class AssetMakeViewSet(ModelViewSet):
+    serializer_class = AssetMakeSerializer
+    queryset = AssetMake.objects.all()
+    permission_classes = [IsAuthenticated, ]
+    authentication_classes = (FirebaseTokenAuthentication,)
+    http_method_names = ['get', 'post']
```
egichuri: could we do also return the asset make here as well instead of id? (for `get`). see comments [here](https://github.com/AndelaOSP/art-backend/pull/80/files)

api/views.py
```diff
@@ -120,3 +120,11 @@ class AssetTypeViewSet(ModelViewSet):
     permission_classes = [IsAuthenticated, ]
     authentication_classes = (FirebaseTokenAuthentication,)
     http_method_names = ['get', 'post']
+
+
+class AssetMakeViewSet(ModelViewSet):
+    serializer_class = AssetMakeSerializer
+    queryset = AssetMake.objects.all()
+    permission_classes = [IsAuthenticated, ]
+    authentication_classes = [FirebaseTokenAuthentication]
```
alexkiura: Remember to add a trailing comma here like so `authentication_classes = [FirebaseTokenAuthentication, ]`

PR #87, opened 2018-05-21
feat(admin view assets): Admin can view all assets via API

Sharonsyra:
### What does this PR do?
• Allow admin to view all assets via API

### Description of tasks to be completed
• Add functionality to allow admin to view all assets
• Add tests

• `python manage.py test`
• Using this URL `http://127.0.0.1:8000/api/v1/assets/` on Postman. Use a Superuser token to view all assets

### What are the relevant pivotal tracker stories
• [#157682134](https://www.pivotaltracker.com/story/show/157682134)

### Screenshots
Admin User Token

screen shot 2018-05-21 at 17 51 41

User Token

screen shot 2018-05-21 at 17 51 56

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -35,7 +35,10 @@ class AssetViewSet(ModelViewSet):
 
     def get_queryset(self):
         user = self.request.user
-        return Asset.objects.filter(assigned_to=user)
+        is_admin = self.request.user.is_staff
+        if not is_admin:
+            return Asset.objects.filter(assigned_to=user)
+        return Asset.objects.all()
 
     def get_object(self):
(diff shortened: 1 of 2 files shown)


api/views.py
```diff
@@ -35,19 +35,19 @@ class AssetViewSet(ModelViewSet):
 
     def get_queryset(self):
         user = self.request.user
+        try:
+            admin = self.request.user.is_staff
```
alexkiura: Since this is a boolean, I think a more descriptive name would be `is_admin`.

Sharonsyra: aah on it.

PR #90, opened 2018-05-23
[#157183717] User can submit an incident report

nelmandela:
### What does this PR do?

  •Allows users to submit an incident report

  •Add an AssetIncidentReport model
  •Make migrations
  •Add test

  • *python manage.py test*
  • `/admin/core/assetincidentreport/`

### Screenshot
incident2
incident
menu


iankigen:
@egichuri @alexkiura @andela-Taiwo Kindly review

core/models/asset.py
```diff
@@ -247,11 +255,32 @@ def save(self, *args, **kwargs):
         super(AssetCondition, self).save(*args, **kwargs)
 
 
+class AssetIncidentReport(models.Model):
```
egichuri: i think we might need to use `TextField ` for some of these fields.. attaching the form used on the app.

egichuri:
screen shot 2018-05-28 at 17 35 20

core/models/asset.py
iankigen: working on it

PR #96, opened 2018-06-03
#157426958 Implements asset filter by email address

JaphethCA:
## What does this PR do?
Implement asset filtering/searching by user email address

## Description of task to be completed
- adds support for email filter via 'email' query param
- test asset filtering with email

## How can this be manually tested?
- python manage.py test
- /api/v1/assets?email=[email_address]

[#157426958](https://www.pivotaltracker.com/story/show/157426958)

## ScreenShot
screen shot 2018-06-03 at 8 01 46 pm
screen shot 2018-06-03 at 7 59 17 pm
screen shot 2018-06-03 at 8 00 17 pm

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -1,4 +1,6 @@
 from django.contrib.auth import get_user_model
+from django.core.validators import validate_email, ValidationError
+from rest_framework import serializers
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
 from rest_framework import status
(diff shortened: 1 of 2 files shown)


api/views.py
```diff
@@ -40,9 +40,17 @@ class AssetViewSet(ModelViewSet):
...
+        queryset = Asset.objects.all()
+        query_params = self.request.query_params
+
         if not is_admin:
-            return Asset.objects.filter(assigned_to=user)
-        return Asset.objects.all()
+            queryset = Asset.objects.filter(assigned_to=user)
+
+        if query_params.get('email'):
+            email = query_params['email']
```
alexkiura: @JaphethCA Looks good but it would be nice to add validation for the email before using it in the query.
You can use the django email validator. You could do something like:
```
from django.core.validators import validate_email
from django.core.exceptions import ValidationError
from rest_framework import serializers

def get_queryset(self):
    .......
    email = query_params['email']
    try:
        validate_email(email)
    except ValidationError as error:
        raise serializers.ValidationError(error.message)
    queryset = Asset.objects.filter(assigned_to__email=email)
```

JaphethCA: Okay, thanks. Will work on that

JaphethCA: Done!

PR #102, opened 2018-06-11
[#155922724] User can import asset makes into the system

iankigen:
#### What does this PR do?
- Allows users to import asset makes

- add asset make import script

-  Run `python post_asset_makes.py`
- Follow the instruction in the terminal

- [#155922724](https://www.pivotaltracker.com/n/projects/2146417/stories/155922724)

#### Screenshots 
screen shot 2018-06-11 at 10 43 20

### DIFF
--- a/assetmake.csv
+++ b/assetmake.csv
@@ -1,0 +1,5 @@
+make_label,asset_type
+Sades,ASSET-TYPE
+Samsung,ASSET-TYPE
+Sony,ASSET-TYPE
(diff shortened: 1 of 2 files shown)


andela-mabdussalam:
Looks good. Could we add an error message if the file does not exist or if the extension is included. This may be useful. 
https://github.com/AndelaOSP/art-backend/blob/develop/import_util.py#L4

PR #108, opened 2018-06-21
#158254542 Import file from url 

JaphethCA:
### What does this PR do?
Implement csv file import from url

### Description of tasks to be completed
- add is_valid_url function to validate url
- add get_csv_from_url function to get csv from url

### What are the relevant pivotal tracker stories
• [#158254542](https://www.pivotaltracker.com/story/show/158254542)

### Screenshots
screen shot 2018-06-21 at 12 43 57 pm

### DIFF
--- a/utils/helpers.py
+++ b/utils/helpers.py
@@ -1,3 +1,9 @@
+import urllib3
+import urllib.parse
+
+urllib3.disable_warnings()
+http = urllib3.PoolManager()
+
 
 def is_valid_file(file_name):
     if file_name.endswith('.csv'):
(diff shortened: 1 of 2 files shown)


utils/helpers.py
```diff
@@ -43,28 +49,41 @@ def display_skipped(result):
...
+    :param filepath: base path for csv files
+    :return: None or file object
+    """
 
-    def get_dependency(self, column):
-        if column in self.columns:
-            colunm_index = self.columns.index(column)
-            if colunm_index > 0:
-                return self.columns[colunm_index - 1]
+    filename = 'url_sample.csv'
```
collinmutembei: Instead of returning a hardcoded value for the filename for all csv files fetched over http, return the actual filename for the file that has been downloaded. That way, if say a user tries to seed data from multiple remote files a local copy of each is created without necessarily overwriting that one file (`url_sample.csv`).

utils/seed_asset.py
```diff
@@ -19,18 +22,28 @@
 
 
 if __name__ == '__main__':
-    filename = input('Enter csv filename (without .csv): ').strip()
+    filename = input('Enter csv filename '
```
collinmutembei: Rename `filename` to be more descriptive, considering the url use case.

JaphethCA: Okay, will do that.

utils/helpers.py
JaphethCA: On it. Thanks for the feedback.

collinmutembei:
LGTM.

PR #110, opened 2018-06-26
#158254668 Add help text to import script

JaphethCA:
## What does this PR do?
- It adds help text to import script

## Description of Task to be completed
- adds help text support with `argparse`

[#158254668](https://www.pivotaltracker.com/story/show/158254668)

## Screenshots
screen shot 2018-06-28 at 10 01 18 am

### DIFF
--- a/utils/seed_asset.py
+++ b/utils/seed_asset.py
@@ -1,6 +1,6 @@
 import sys
 import os
-
+import argparse
 import django
 
 project_dir = os.path.dirname(
@@ -22,6 +22,14 @@ django.setup()
 
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        prog="seed_asset",
+        description="""
+            The program for importing of assets
+        """
+    )
+    args = parser.parse_args()
+
     filename_or_url = input('Enter csv filename '
                             '(without .csv or url link to csv): ').strip()
     filepath = os.path.abspath(os.path.join(os.path.dirname(__file__))) + '/'  # noqa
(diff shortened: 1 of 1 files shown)


utils/seed_asset.py
```diff
@@ -22,6 +22,14 @@
...
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        prog="seed_asset",
+        description="""
+            The program for importing of assets
+        """
+    )
+    args = parser.parse_args()
+
```
egichuri: @JaphethCA `if not show the message and continue with execution.` - is this possible?

utils/seed_asset.py
```diff
@@ -22,6 +22,14 @@
 
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        prog="seed_asset",
+        description="""
+            The program for importing of assets
```
egichuri: we should include the expected column names here.

egichuri: and file format, etc

utils/seed_asset.py
JaphethCA: Am sorry, but I don't understand what you mean. Can you please give more details.

utils/seed_asset.py
JaphethCA: Okay, Will do that. Thanks

utils/seed_asset.py
egichuri: currently the `-h` argument shows the help text and exits.. without the `-h` it goes straight to execution (asks user to enter the csv file name) without displaying the help text. is it possible to show the help text even without the `-h`?

JaphethCA: Sure, but how will user run the command to input csv file or link if running the script without arguments return a help text?

PR #120, opened 2018-07-05
#156802464 Implement  continuous  deployment

cjmash:
**What does this PR do?**

 - Creates a continuous integration and deployment pipeline 

-  Breaks down the pipeline into three jobs namely build, tests and deployment

**Description of tasks to be completed**

• Once features are merged into the master or the develop  branch, A deployment is triggered 
  that will push the new features to GCP staging or production

**What are the relevant pivotal tracker stories**

[#156802464](https://www.pivotaltracker.com/story/show/156802464)

### DIFF
--- a/settings/dev.py
+++ b/settings/dev.py
@@ -1,5 +1,5 @@
 from .base import *  # noqa: F403,F401
-
+import os
 DEBUG = True
 
 INSTALLED_APPS += [  # noqa ignore=F405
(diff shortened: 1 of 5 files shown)


settings/dev.py
```diff
@@ -10,12 +10,7 @@
     'debug_toolbar.middleware.DebugToolbarMiddleware',
 ]
 
-ALLOWED_HOSTS = [
-    'api-staging-art.andela.com',
-    '0.0.0.0',
-    '127.0.0.1',
-    'art-backend.herokuapp.com'
-]
+ALLOWED_HOSTS = [os.environ.get('MY_HOST_IP'), 'api-staging-art.andela.com']
```
egichuri: @cjmash will we need this for the prod env as well?

cjmash: yeah @egichuri , I have updated the prod file

egichuri: great!

scott45:
Good work Collins. I realise your deployment jobs have been named explicitly (staging and deployment). Environments are already catered for in the k8's script. I suggest you use one deployment job and then filter the 2 branches that should trigger deploy.sh script.  It'll give us the same result with minimal code. 

What does this function do in the deploy.sh script?
https://github.com/AndelaOSP/art-backend/blob/41edc6df4987f3c27c21dbc587186754dc431300/scripts/deploy.sh#L22

PR #122, opened 2018-07-09
Feat pagination enhancement #158792993

precious-ijege:
**What does this PR do?**

- It allows custom  pagination on API endpoints using **page_size** and **page** as parameters.

- By default, pagination is set to True, hence the response from data is paginated with a default 
**page_size**.

- If **paginate** parameter is set to false, every data for that endpoint is returned.

- You can also explicitly set the value of **page_size** in the url.

- It also allows you toggle between pages by explicitly setting the **page** parameter in the url.

**Description of Task to be completed?**

- Add  the new custom pagination class to the app settings.

- Ensure all tests are passing.

**How should this be manually tested?**
- Visit any of the available endpoints.

**Screenshot**

screen shot 2018-07-13 at 2 22 38 pm
screen shot 2018-07-13 at 2 24 08 pm
screen shot 2018-07-13 at 2 24 40 pm


QUDUSKUNLE:
@precious-ijege  `api/tests/test_asset_condition_api.py` is conflicting, kindly correct that.

egichuri:
@precious-ijege please fix the conflicts and update your branch with develop

core/pagination.py
```diff
@@ -0,0 +1,5 @@
+from rest_framework.pagination import PageNumberPagination
+
+
+class PageNumberPagination(PageNumberPagination):
+    page_size_query_param = 'limit'
```
collinmutembei: A more descriptive name would be `page_size`. Using `limit` could lead to confusion; since the same value is the default query parameter for [`rest_framework.pagination.LimitOffsetPagination`](http://www.django-rest-framework.org/api-guide/pagination/#limitoffsetpagination)

collinmutembei: We can still limit the number of items per page my setting
```python
page_size = 10
```
The effect of this would be
 - All API responses return 10 items per page
- API requests that require the entire response can specify a greater `page_size` param i.e `?page_size=999`

This approach ensures that the more frequent API calls i.e `/api/v1/assets` do not fetch all the data while allowing for the FE to be explicit when requesting for all data.

egichuri: I think we could have a different way to specify all data, that a number.. if it gets to a point where we have eg more that 999 assets it'd still be paginated.
Is it possible to add another parameter? something like `paginated=false` which defaults to `true` if not specified?

core/pagination.py
```diff
@@ -0,0 +1,35 @@
...
+        if ret < 0 or (ret == 0 and strict):
+            raise ValueError()
+        if cutoff:
+            return min(ret, cutoff)
+        return ret
+
+
+class PageNumberPagination(PageNumberPagination):
+    page_size_query_param = 'page_size'
+    page_size = 3
```
collinmutembei: Update the default `page_size` to 20

PR #123, opened 2018-07-09
#158853569 Create asset specs model

JaphethCA:
## What does this PR do?
- Creates AssetSpecs model 

## Description of tasks to be completed
- write tests for model
- adds AssetSpecs to admin panel

[#158853569](https://www.pivotaltracker.com/story/show/158853569)

## Screenshot
screen shot 2018-07-09 at 2 57 13 pm


core/models/asset.py
```diff
@@ -173,6 +240,10 @@ class Asset(models.Model):
     asset_condition = models.CharField(editable=False,
                                        max_length=50,
                                        default='Brand New')
+    specs = models.ForeignKey(AssetSpecs,
+                              blank=True,
+                              null=True,
+                              on_delete=models.CASCADE)
```
egichuri: let's use `PROTECT` here..

egichuri: while we're here.. please change https://github.com/AndelaOSP/art-backend/pull/123/files#diff-bea1207c116ad66b7b398915447bd091R129 to `PROTECT` as well

JaphethCA: Done! Thanks for the feedback.

core/models/asset.py
```diff
@@ -399,3 +470,27 @@ def allocation_history_post_save(sender, **kwargs):
             current_status=ALLOCATED
         )
         asset_status.save()
+
+
+@receiver(pre_save, sender=AssetSpecs)
```
egichuri: do we need this validation when using the `choices` parameter?

egichuri:
this is a test (ignore)

PR #124, opened 2018-07-11
#15886823 Admin add asset purchase date

QUDUSKUNLE:
### What does this PR do?
- It allows an admin to add assets purchase date 

- Add purchase date column to Asset model
- Ensure no test is broken

- Go to admin dashboard and add new asset

### Screenshot
- 
screen shot 2018-07-11 at 2 27 39 am
screen shot 2018-07-11 at 2 28 25 am


core/migrations/0011_auto_20180710_1540.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.0.1 on 2018-07-10 15:40
```
collinmutembei: Looks like you generated several migrations over the course on adding the `purchase_date` field. While checking all of them in still works fine at the system level, I have often found that generating one migration file for every change on the models results in a more readable diff. Having one file to review when there are changes affecting the database means that the team has a better understanding of the implication of the changes. You could delete<sup>1</sup> the three migration files (`0009_asset_purchase_date.py`, `0010_auto_20180710_1244.py` and `0011_auto_20180710_1540.py`) and generate one consolidated file.

1. **Remember** to commit the deletions

core/models/asset.py
```diff
@@ -231,6 +231,7 @@ class Asset(models.Model):
     serial_number = models.CharField(
         unique=True, null=True, blank=True, max_length=50)
     created_at = models.DateTimeField(auto_now_add=True, editable=False)
+    purchase_date = models.DateField()
```
collinmutembei: Considering the nature of the purchase date; It **cannot** be in the future...how about we include a validator method to the `purchase_date` model field to raise a _ValidationError_ when the date is set to a value in the future.
I do realise that validation that this would increase the scope of the task. What do you think @QUDUSKUNLE @egichuri ?

core/migrations/0011_auto_20180710_1540.py
QUDUSKUNLE: @collinmutembei thanks, changes effected.

core/models/asset.py
egichuri: since it's validation for a field added it can be done as part of this ticket

collinmutembei: @QUDUSKUNLE You can reference Django's [docs](https://docs.djangoproject.com/en/2.0/ref/validators/) to understand how to go about it.

QUDUSKUNLE: @collinmutembei Thanks. Changes effected.

core/validator.py
```diff
@@ -0,0 +1,10 @@
+import datetime
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext_lazy as _
+
+
+def validate_date(value):
+    if value > datetime.date.today():
+        raise ValidationError(
+            _('%(value)s invalid purchase date'), params={'value': value},
```
collinmutembei: As an admin it might to be clear on the onset as to what makes a purchase date invalid. Update this message to be more convey the meaning. Thanks.

QUDUSKUNLE: Can you suggest? Or is this okay `Purchase date cannot be in the future`

collinmutembei: That sounds better @QUDUSKUNLE. It does leave out one crucial part though; what a valid purchase date it. There's no harm in being verbose as necessary.

collinmutembei: > "2018-07-02 is invalid. Purchase date can only be set to today or a date in the past"

PR #132, opened 2018-07-17
[Finishes #158909217] Fix asset subcategory

andela-Taiwo:
#### What does this PR do
- Fix bug asset subcategory does not exist error

#### Description of tasks completed
- add logic to filter existing subcategory
- add logic to get an instance of subcategory

#### Related PivotalTracker
- #158909217


utils/helpers.py
```diff
@@ -106,7 +106,7 @@ def write_record_skipped(record, file_path):
     :return: None
     """
     filename = "/skipped.csv"
-    fieldnames = ('Count', 'Make', 'Type', 'Asset Code', 'Category',
+    fieldnames = ('count', 'Make', 'Type', 'Asset Code', 'Category',
```
collinmutembei: All the field names are in title case. What's the reasoning behind changing the "count" field name to lowercase?

utils/asset/post_scripts.py
```diff
@@ -400,7 +400,8 @@ def post_asset_types(f, file_length): # noqa
                 else:
                     asset = AssetType()
                     asset.asset_type = asset_type
-                    asset.asset_sub_category = sub_category_name
+                    asset.asset_sub_category = AssetSubCategory.objects\
```
collinmutembei: This query is redundant considering that we have an AssetSubCategory model object called `sub_category_name` already.
https://github.com/AndelaOSP/art-backend/blob/de8ce2b9824577cf4c5c5b95478efb3ae744e01d/utils/asset/post_scripts.py#L386-L387
Assigning `asset.asset_sub_category` to it is sufficient.
```python
asset.asset_sub_category = sub_category_name
```

utils/helpers.py
andela-Taiwo: It was conflicting with the field names in the dictionary.

collinmutembei:
@andela-Taiwo Add tests for the import functionality.

utils/helpers.py
collinmutembei: What dictionary @andela-Taiwo?

collinmutembei:
@andela-Taiwo @kmazi  Are you blocked on tests?

PR #135, opened 2018-07-23
Enables user to add and view an office block

esirK:
#### What does this PR do?
- It enables a user to add and view an office block

- Create a building model
- Add test for the building model
- Go to admin dashboard and add a new building
#### Relevant pivotal tracker stories
[#159022585](https://www.pivotaltracker.com/story/show/159022585)
#### Screenshot
- 
screen shot 2018-07-23 at 14 03 14


core/models/building.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+
+
+class Building(models.Model):
```
collinmutembei: Let's refer to the model as `OfficeBlock` rather than `Building`. Pardon that this was not stated clearly on the acceptance criteria. Should be a minor refactor.

core/migrations/0015_auto_20180723_1033.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.0.1 on 2018-07-23 10:33
```
collinmutembei: Ensure that you have only one migration file after the refactor.

core/models/building.py
collinmutembei: Also, while at it...rename the model field from `block_name` to `name`. See number *15* [here](https://goo.gl/6gqMkb) to understand why.

collinmutembei:
@egichuri To avoid a scenario where multiple entries are referring the same office block, we might need to enforce the case sensitivity as we did on [other](https://github.com/AndelaOSP/art-backend/blob/develop/core/models/asset.py#L86-L88) unique fields.

screen shot 2018-07-23 at 14 41 45

api/serializers.py
```diff
@@ -376,3 +375,9 @@ def validate(self, fields):
                 "Similar asset specification already exist"
             )
         return fields
+
+
+class BuildingSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Building
+        fields = ("block_name", )
```
egichuri: I think we should include the `id` as well

egichuri:
@collinmutembei yes.. let's do that. @esirK see the comment above

esirK:
Working on the comments right now @egichuri

api/serializers.py
collinmutembei: A custom `id` field?
Remember Django adds an `id` field as the primary_key by [default](https://docs.djangoproject.com/en/2.0/topics/db/models/#automatic-primary-key-fields).

esirK: @egichuri Is it inadvisable to depend on Django's default `id` field??

egichuri: it's ok to use the default id.. but let's expose it on fields available to the API

esirK: Got it 👍

esirK:
@collinmutembei on 

> To avoid a scenario where multiple entries are referring the same office block, we might need to enforce the case sensitivity as we did on other unique fields.
I realized we are using something like `self.name.title()` but this does not prevent a user from enterling a name like `Block A` then `Block    A`.
I think using something like `" ".join(self.name.title().split())` might help. What do you think??

egichuri:
@esirK since this happens before the object is saved the value that goes to the db will always be `Block A` (for the above). The db will then enforce the unique constraint


p

esirK:
@egichuri  I am saying something like `Block A` and `Block       A` notice the double space in the second case.

egichuri:
oh.. hadn't noticed that. @collinmutembei do we need to worry about that? if so we'd need to update the other places as well

collinmutembei:
Nice catch @esirK, considering that the goal of doing so was to enforce uniqueness across different cases of the user input...this does go the extra mile of invalidating any whitespace that could have gotten around the initial implementation. @egichuri We can work with this, then update all other section later.

esirK:
@collinmutembei @egichuri We have worked on the requested changes.

egichuri:
awesome!

collinmutembei:
screen shot 2018-07-23 at 18 59 41

core/models/officeblock.py
```diff
@@ -0,0 +1,15 @@
+from django.db import models
+
+
+class OfficeBlock(models.Model):
+    name = models.CharField(max_length=50,
+                            blank=False, null=False, unique=True)
+
+    def clean(self):
+        self.name = " ".join(self.name.title().split())
```
egichuri: For this to work you'll need to override `.save` and call `self.full_clean()`

esirK: Thanks, @egichuri  We had missed this.
Done.

PR #137, opened 2018-07-23
#159023098 Modify asset specs addition

Japhethca:
## What does this PR do?
- Implement creating asset specs via asset endpoint

## Description of the tasks to be completed
- modify `AssetSerializer`'s `to_internal_values` serializer method to include asset specs
- test for specs creation via asset endpoint

[#159023098](https://www.pivotaltracker.com/story/show/159023098)

## Screenshot
screen shot 2018-07-23 at 10 25 05 pm
screen shot 2018-07-23 at 10 25 32 pm

### DIFF
--- a/api/tests/test_manage_asset_api.py
+++ b/api/tests/test_manage_asset_api.py
@@ -134,10 +134,7 @@ class ManageAssetTestCase(APIBaseTestCase):
             "asset_code": "IC002",
             "serial_number": "SN002",
             "model_number": self.assetmodel.model_number,
-            "purchase_date": "2018-07-10",
-            "processor_type": "Intel core i3",
-            "processor_speed": 2.3,
-            "screen_size": 15
+            "purchase_date": "2018-07-10"
         }
         response = client.post(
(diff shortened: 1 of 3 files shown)


api/tests/test_manage_asset_api.py
```diff
@@ -134,7 +134,10 @@ def test_admin_can_post_asset(self, mock_verify_id_token):
             "asset_code": "IC002",
             "serial_number": "SN002",
             "model_number": self.assetmodel.model_number,
-            "purchase_date": "2018-07-10"
+            "purchase_date": "2018-07-10",
+            "processor_type": "Intel core i3",
+            "processor_speed": 2.3,
+            "screen_size": 15
         }
```
egichuri: i think we should also have a test for invalid data. eg a screen size not present in the set choices.

egichuri: i think we should also have a test for invalid data. eg a screen size not present in the set choices.
And a test to confirm an asset is created when the specs aren't provided.

PR #144, opened 2018-08-02
[#159647851]: Adds a department 

esirK:
### What does this PR do?
- It adds a Department

-Adds Department model
-Adds DepartmentSerializer
-Adds Department ViewSet
-Adds Department Model tests
-Adds Department Api tests
-Adds Department Url

- Test out the `departments` endpoint with admin right

### Relevant Stories
- [#159022596](https://www.pivotaltracker.com/story/show/159022596)

### Screenshot
screen shot 2018-08-02 at 12 52 37


api/tests/test_department_api.py
```diff
@@ -0,0 +1,95 @@
...
+    def test_editing_department(self,
+                                mock_verify_id_token):
+        mock_verify_id_token.return_value = {'email': self.admin.email}
+        name = "People"
+        data = {"name": name}
+        client.post(
+            self.department_url,
+            data=data,
+            HTTP_AUTHORIZATION="Token {}".format(self.token_user))
+        department_id = Department.objects.get(name=name).id
```
esirK: Hey @collinmutembei  for this test, we manually queried the `department id` since the `auto increment` id field in the test database does not reset which we don't yet understand why since on other tests it gets reset and we can be sure of the id number, is this the right approach or is there a way to ensure that the field gets reset automatically??

collinmutembei: @esirK The goal here is to test that the edit functionality provided through PUT works as expected:
- Given an existing department, updates on the name should be reflected
- Only authenticated admin users can edit a department's name

That being said; it's not clear to me why the `id` needs to reset. Find me, so we can discuss this.

core/models/department.py
```diff
@@ -0,0 +1,27 @@
...
+    def save(self, *args, **kwargs):
+        try:
+            self.full_clean()
+        except Exception as e:
+            raise ValidationError(e)
+        super().save(*args, **kwargs)
+
+    class Meta:
+        verbose_name = "Department"
+        unique_together = ("name", )
```
collinmutembei: Having one field for [`unique_together`](https://docs.djangoproject.com/en/2.0/ref/models/options/#unique-together) defeats the purpose. See docs.

collinmutembei:
@esirK is this still a WIP?

esirK:
@collinmutembei yes. We are currently working on the second part for allowing an asset to be assigned to a department

esirK:
Hey @collinmutembei, I have Updated this PR to just be about adding a Department is this alright with you? We can then work on assigning an asset to a department as its own task.

collinmutembei:
Good idea @esirK. We should have split the tasks earlier on. Update PT as well.

core/models/department.py
```diff
@@ -0,0 +1,26 @@
...
+    last_modified = models.DateTimeField(auto_now=True, editable=False)
+
+    def clean(self):
+        self.name = " ".join(self.name.title().split())
+
+    def save(self, *args, **kwargs):
+        try:
+            self.full_clean()
+        except Exception as e:
+            raise ValidationError(e)
```
collinmutembei: Raise a DRF exception here

screen shot 2018-08-09 at 12 00 21

collinmutembei: @esirK I could be wrong as to whether you need to raise the DRF exception in the model. Let's close this soonest, so that we can unblock the subsequent tasks that depend on these changes.

collinmutembei: Ah! this solves it.
```python
from rest_framework import serializers
...

except ValidationError as error:
            raise serializers.ValidationError(error)
```

esirK: So no changes need to be implemented right?? @collinmutembei

PR #155, opened 2018-08-08
#159594130 Model clean up

collinmutembei:
- [X] Remove redundant fields on **Asset** model.
- [X] Refactor `post_save` signals.


collinmutembei:
@esirK Carry on from here.

PR #159, opened 2018-08-13
#159022596 Allow an asset to be assigned to a department

esirK:
### What does this PR do?

- It allows an asset to be assigned to a department

### Description of task completed.

- write `AssetAssignee` model 
- Write post_save signals to save `AssetAssignee` when a `User` or a `Department` is saved
- fix tests to use new AssetAssignee as the `assigned_to`

- visit the **allocations** endpoint with admin right and try to allocate an asset to a department

- Run tests to ensure nothing broke

### Screenshots
screen shot 2018-08-14 at 01 57 53


core/models/asset.py
```diff
@@ -482,27 +516,76 @@ def save_initial_asset_status(sender, **kwargs):
...
+def assetassignee_user(sender, instance, created, **kwargs):
+    if created or not hasattr(instance, 'assetassignee'):
+        AssetAssignee.objects.create(user=instance)
+    instance.assetassignee.save()
+
+
+@receiver(post_save, sender=Department)
+def assetassignee_department(sender, instance, created, **kwargs):
+    if created or not hasattr(instance, 'assetassignee'):
+        AssetAssignee.objects.create(department=instance)
```
collinmutembei: Add the statement below at the end of this `post_save` signal receiver to ensure a **Department** object reference on the **AssetAssignee** model is updated if it is updated.
```python
instance.assetassignee.save()
```

core/models/asset.py
```diff
@@ -301,6 +302,42 @@ class Meta:
...
+
+    def __str__(self):
+        if self.department:
+            return str(self.department)
+        if self.user:
+            return str(self.user)
+        else:
+            return "No Assignee Here"
+
+    @property
```
collinmutembei: Creating `@property` methods on the **AssetAssignee**  model to resolve fields on the **User**/**Department** model works for now, though it's not scalable. A better approach (dare I say _DRY_ er) is to reference the respective fields on the **User**/**Department**  model themselves. All the same, let's keep it as is for now. We'll work on refactoring this at a later stage.

core/models/asset.py
```diff
@@ -301,6 +302,42 @@ class Meta:
...
+    def last_name(self):
+        if self.department:
+            return self.department.name
+        if self.user:
+            return self.user.last_name
+
+    @property
+    def email(self):
+        if self.department:
+            return self.department.name
```
collinmutembei: **Note to self**: Department email?

screen shot 2018-08-14 at 13 13 32

api/tests/test_allocation_api.py
```diff
@@ -99,7 +100,7 @@ def test_asset_status_changes_to_allocated(self, mock_verify_id_token):
         mock_verify_id_token.return_value = {'email': self.user.email}
         data = {
             'asset': self.asset.id,
-            'current_owner': self.user.id}
+            'current_owner': self.asset_assignee.id}
         token = f"Token {self.token_user}"
         response = client.post(
             self.allocations_urls, data,
```
collinmutembei: Add test(s) for allocation to a department.

collinmutembei:
Also, Update the PR title to match the recommended convention [here](https://github.com/andela/bestpractices/wiki/Git-naming-conventions-and-best-practices#pr-naming). Follow the commit message format as well next time.

core/models/asset.py
esirK: on this, what would you suggest? Should we add an email field to a department or return an empty string??

collinmutembei: Leave it as is for now.

PR #160, opened 2018-08-14
#159759006  Implement gunicorn gevent worker 

cjmash:
**What does this PR do?**
Adds the gevent worker to the requirements file.
This will enable the application to start with gevent gunicorn class.

**Description of task completed.**
    - Use gunicorn async classes to prevent the workers timeout issue that was causing a 502 error.


collinmutembei:
@cjmash Resolve the conflict in `requirements.txt`

PR #167, opened 2018-08-27
#159994135 Filter assets  by model numbers and asset type via API.

esirK:
### What does this PR do?
- It allows filtering of assets by querying using `model_number`, `asset_type` or `email`

-Adds model_number filter
-Adds asset_type filter
-Adds an email filter

- Test out the `assets` endpoint with admin right

### Relevant Stories
- [#159994135](https://www.pivotaltracker.com/story/show/159994135)

### Screenshot
screen shot 2018-08-27 at 17 21 35
screen shot 2018-08-27 at 17 22 20
screen shot 2018-08-27 at 17 22 08
screen shot 2018-08-27 at 17 23 49
screen shot 2018-08-27 at 17 24 08
screen shot 2018-08-27 at 17 27 48

screen shot 2018-08-27 at 17 28 06
screen shot 2018-08-27 at 17 29 20

### DIFF
--- a/api/filters.py
+++ b/api/filters.py
@@ -1,0 +1,31 @@
+from django_filters import rest_framework as filters
+
+from core.models import Asset
+
+
+class AssetFilter(filters.FilterSet):
+    email = filters.CharFilter(
+        field_name='assigned_to__user__email',
+        lookup_expr='icontains', )
+    model_number = filters.CharFilter(
+        field_name='model_number__model_number',
+        lookup_expr='iexact',
+        method='filter_with_multiple_query_values')
+    asset_type = filters.CharFilter(
+        field_name='model_number__make_label__asset_type__asset_type',
+        lookup_expr='iexact',
+        method='filter_with_multiple_query_values')
+
+    def filter_model_number(self, queryset, name, value):
+        return queryset.filter(model_number__model_number__iexact=value)
+
+    def filter_asset_type(self, queryset, name, value):
+        return queryset.filter(model_number__make_label__asset_type__asset_type__iexact=value)
+
+    def filter_with_multiple_query_values(self, queryset, name, value):
+        lookup = '__'.join([name, 'in'])
+        return queryset.filter(**{lookup: value.split(',')})
+
+    class Meta:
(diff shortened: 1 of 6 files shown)


api/views.py
```diff
@@ -80,30 +81,60 @@ def create(self, request, *args, **kwargs):
...
+            validate_email(value)
+            user_asset_assignee = User.objects.get(email=value)
+            queryset = Asset.objects.filter(assigned_to__user=user_asset_assignee)
+            for query, value in query_params.items():
+                if query == 'email':
+                    pass
+                else:
+                    queryset = _filters[query](queryset, name, value)
+        except (ValidationError, ObjectDoesNotExist):
+            queryset = Asset.objects.none()
```
esirK: Hello @collinmutembei, @egichuri Previously, a `ValidationError` was being raised at this point. Should this still be the case or we can just return an empty queryset? If we raise a  `ValidationError` here, does it mean we should raise the error even when an `asset_type` or `model_number` does not exist instead of returning empty lists??

egichuri: @esirK was the validation error from `validate_email(value)`?

esirK: Yes it was @egichuri .

egichuri: you could catch the exceptions separately
```
try:
    # as it is
except ValidationError:
    # email validation error
except ObjectDoesNotExist:
    # empty queryset
```

api/filters.py
```diff
@@ -0,0 +1,31 @@
...
+    model_number = filters.CharFilter(
+        field_name='model_number__model_number',
+        lookup_expr='iexact',
+        method='filter_with_multiple_query_values')
+    asset_type = filters.CharFilter(
+        field_name='model_number__make_label__asset_type__asset_type',
+        lookup_expr='iexact',
+        method='filter_with_multiple_query_values')
+
+    def filter_model_number(self, queryset, name, value):
```
collinmutembei: This method is redundant, as the statements on L11 and L12 would result in the same queryset. Delete `filter_model_number()` and `filter_asset_type()` below.

PR #171, opened 2018-09-03
#160137335 Admin filter users

QUDUSKUNLE:
### What does this PR do?
- It allows the admin to filter users by `cohort` or`email` or `asset_count`

- Add UserFilter class to filter module
- Write tests for admin filter users

- Test out the `users` endpoint with admin right and query by `cohort` or `email` or  `asset_count`

### Screenshot
- 
screen shot 2018-09-03 at 4 03 13 pm

screen shot 2018-09-03 at 4 05 28 pm

screen shot 2018-09-03 at 4 05 59 pm

### DIFF
--- a/api/filters.py
+++ b/api/filters.py
@@ -32,19 +32,24 @@ class UserFilter(filters.FilterSet):
 
     email = filters.CharFilter(
         field_name='email',
-        lookup_expr='icontains',)
+        lookup_expr='istartswith',
+        method='filter_by_email_or_allocated_asset_count',)
 
     asset_count = filters.CharFilter(
         field_name='allocated_asset_count',
         lookup_expr='iexact',
-        method='filter_with_multiple_query_values',)
+        method='filter_by_email_or_allocated_asset_count',)
 
-    def filter_with_multiple_query_values(self, queryset, name, value):
-        users = [
-            user.id
-            for user in queryset
-            if user.assetassignee.current_owner_asset.count() == int(value)]
-        return User.objects.filter(id__in=users)
+    def filter_by_email_or_allocated_asset_count(self, queryset, name, value):
+        if name == 'email':
+            return queryset.filter(email__istartswith=value)
+        else:
+            users = [
+                user.id
+                for user in queryset
+                if user.assetassignee.current_owner_asset.count() == int(value)
+            ]
+            return User.objects.filter(id__in=users)
 
     class Meta:
         model = User
--- a/api/filters.py
+++ b/api/filters.py
@@ -32,24 +32,21 @@ class UserFilter(filters.FilterSet):
 
     email = filters.CharFilter(
         field_name='email',
-        lookup_expr='istartswith',
-        method='filter_by_email_or_allocated_asset_count',)
+        lookup_expr='istartswith',)
 
     asset_count = filters.CharFilter(
         field_name='allocated_asset_count',
+        label='Asset count',
         lookup_expr='iexact',
-        method='filter_by_email_or_allocated_asset_count',)
+        method='filter_by_allocated_asset_count',)
 
-    def filter_by_email_or_allocated_asset_count(self, queryset, name, value):
-        if name == 'email':
-            return queryset.filter(email__istartswith=value)
-        else:
-            users = [
-                user.id
-                for user in queryset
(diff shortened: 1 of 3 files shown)


api/tests/test_user.py
```diff
@@ -336,7 +363,7 @@ def test_allocated_assets_are_returned(self, mock_verify_token):
             HTTP_AUTHORIZATION="Token {}".format(self.token_admin)
         )
         count = len(response.data['allocated_assets'])
-        self.assertEqual(count, 0)
+        self.assertEqual(count, 1)
```
collinmutembei: This assert statement was initially validating that the count will always be zero for users who do not have an asset allocated to them. To ensure that we keep that check, you could assign `self.asset` to a different user that's not `self.user`. Alternatively, you could move the allocation on [line 47](https://github.com/AndelaOSP/art-backend/pull/171/files#diff-a88fe99b551eef0f172e7505561b4951R47) to within `test_admin_filter_users_by_asset_count()`, where you're validating the filtering by asset count.

api/tests/test_user.py
```diff
@@ -347,5 +374,62 @@ def test_allocated_assets_are_returned(self, mock_verify_token):
...
+        self.assertEqual(response.data['results'][0]['cohort'],
+                         self.user.cohort)
+        self.assertEqual(response.data['count'], 2)
+        self.assertEqual(response.status_code, 200)
+
+    @patch('api.authentication.auth.verify_id_token')
+    def test_admin_user_filter_users_by_email(self, mock_verify_token):
+        mock_verify_token.return_value = {'email': self.admin_user.email}
+        response = client.get(
+            '{}?email={}'.format(self.users_url, self.user.email),
```
collinmutembei: Refactor tests related to filtering by email to emulate a letter in the alphabet as the passed value (case insensitive). 👇 

screen shot 2018-09-03 at 21 03 28

api/tests/test_user.py
collinmutembei: Prefer the latter.

QUDUSKUNLE:
@collinmutembei Requested changes effected.

api/filters.py
```diff
@@ -23,3 +23,34 @@ def filter_with_multiple_query_values(self, queryset, name, value):
...
+
+    email = filters.CharFilter(
+        field_name='email',
+        lookup_expr='istartswith',
+        method='filter_by_email_or_allocated_asset_count',)
+
+    asset_count = filters.CharFilter(
+        field_name='allocated_asset_count',
+        lookup_expr='iexact',
+        method='filter_by_email_or_allocated_asset_count',)
```
collinmutembei: Add a [label](https://django-filter.readthedocs.io/en/latest/ref/filters.html#label) argument to the `asset_count` field with a value of '_exact allocated asset count_'.

api/filters.py
```diff
@@ -23,3 +23,34 @@ def filter_with_multiple_query_values(self, queryset, name, value):
...
+
+class UserFilter(filters.FilterSet):
+    cohort = filters.CharFilter(
+        field_name='cohort',
+        lookup_expr='iexact',)
+
+    email = filters.CharFilter(
+        field_name='email',
+        lookup_expr='istartswith',
+        method='filter_by_email_or_allocated_asset_count',)
```
collinmutembei: Specifying a [method](https://django-filter.readthedocs.io/en/latest/ref/filters.html#method) argument on a field, means you want to resolve the filtering on the queryset yourself rather than leaving `django-filter` to do it for you. It defeats the purpose, since the `field_name` and `lookup_expr` arguments are sufficient to return the filtered result we would like.

api/filters.py
```diff
@@ -23,3 +23,34 @@ def filter_with_multiple_query_values(self, queryset, name, value):
...
+
+    asset_count = filters.CharFilter(
+        field_name='allocated_asset_count',
+        lookup_expr='iexact',
+        method='filter_by_email_or_allocated_asset_count',)
+
+    def filter_by_email_or_allocated_asset_count(self, queryset, name, value):
+        if name == 'email':
+            return queryset.filter(email__istartswith=value)
```
collinmutembei: This statement is redundant since `django-filter` would have already resolved the email field above to the same queryset.

PR #182, opened 2018-09-19
[#160138951] Allow user to import assets from an Api endpoint (WIP)

esirK:
### What does PR do?

- Adds an Api endpoint where a user can upload a csv file with assets details to.

### Description of the task to be completed:

  •Add an Api endpoint where a user can upload a csv file to

  •As an admin, using postman send a csv file to the `uploads` endpoint. If the file contains the 3 required fields **(Asset Code, Serial No., Notes)** , the assets should be saved in the database.

### Relevant pivotal tracker stories
[#160138951](https://www.pivotaltracker.com/story/show/160138951)


api/views.py
```diff
@@ -362,3 +364,47 @@ def destroy(self, request, *args, **kwargs):
...
+                       'screen_size': screen_size,
+                       'processor_type': processor_type,
+                       'storage': storage,
+                       'memory': memory})
+
+        asset_specs = AssetSpecs.objects.get_or_create(**kwargs)
+
+        assets = []
+
+        for pos, line in enumerate(file_obj):
```
egichuri: i think it might be better to use the csv module which takes care of column headers. if a file has the columns ordered differently the implementation here would be wrong. eg if the imported file has serial number in the first column.

api/views.py
```diff
@@ -362,3 +364,47 @@ def destroy(self, request, *args, **kwargs):
...
+            asset_fields = line.decode('UTF-8').rstrip().split(',')
+
+            asset_code = asset_fields[0]
+            serial_number = asset_fields[1]
+            notes = asset_fields[2]
+
+            assets.append(Asset(model_number=model_number, asset_code=asset_code, serial_number=serial_number,
+                                notes=notes, asset_specs=asset_specs))
+        Asset.objects.bulk_create(assets)
+        return Response(status=204)
```
egichuri: how are we handling a failure?

api/views.py
esirK: Noted

api/views.py
esirK: Working on this.

esirK:
Hey, @egichuri Checkout the changes.

egichuri:
@esirK lgtm.. just fix the pep8 issue

PR #199, opened 2018-10-31
[#161544768]:Get assets using serial number or asset code

nelmandela:
### What does this PR do?

 •Allows user to get assets using serial_number or asset_code

  •Add a queryset in the ManageAssetViewSet

  Navigate to assets endpoint `http://127.0.0.1:8000/api/v1/manage-assets/?asset_code=`

### Relevant PT stories:

[#161544768](https://www.pivotaltracker.com/story/show/161544768)

### Screenshots

screen shot 2018-11-06 at 15 17 58


egichuri:
@nelmandela i believe this should be for `/assets/` not `/manage-assets/`

egichuri:
@nelmandela what's the holdup here?

nelmandela:
> @egichuri we were working on a couple of errors we were experiencing but its done now

egichuri:
@nelmandela which errors?

api/views.py
```diff
@@ -90,7 +90,15 @@ class AssetViewSet(ModelViewSet):
...
+        has_filter = False
+        for field in self.request.query_params:
+            if field == 'serial_number' or field == 'asset_code':
+                query_filter[field] = self.request.query_params.get(field)
+                has_filter = True
+        
+        queryset = Asset.objects.filter(assigned_to=asset_assignee, **query_filter) if\
+            has_filter else Asset.objects.filter(assigned_to=asset_assignee)
 
         return queryset
```
egichuri: we can break this instead of using a one-liner for readability.
```
if has_filter:
    return Asset.objects.filter(assigned_to=asset_assignee, **query_filter)

return Asset.objects.filter(assigned_to=asset_assignee)
```

nelmandela:
@egichuri first we were unable to fetch assets using `asset_url` that worked but then it wasn't getting all assets assigned to the logged in user.we've been unable to figure out why that is happening but we discovered its only happening on my end @kmazi works fine

api/views.py
nelmandela: on it

nelmandela:
@egichuri done

api/views.py
```diff
@@ -95,7 +95,18 @@ class AssetViewSet(ModelViewSet):
...
         asset_assignee = AssetAssignee.objects.filter(user=user).first()
-        queryset = Asset.objects.filter(assigned_to=asset_assignee)
+        query_filter = {}
+        has_filter = False
+        for field in self.request.query_params:
+            if field == 'serial_number' or field == 'asset_code':
+                query_filter[field] = self.request.query_params.get(field)
+                has_filter = True
+
+        queryset = Asset.objects.filter(assigned_to=asset_assignee, **query_filter)
```
egichuri: i believe we should include ` **query_filter` only if `has_filter ` is `True`

and 1 last change.. we should allow security users to query this endpoint for all assets. ie if it's a security user making the request they can see assets belonging to other users (we shouldn't check `assigned_to ` for requests from security users)

nelmandela: alright

PR #201, opened 2018-11-06
feat[#161049350]: An admin can edit and delete a security user

precious-ijege:
### What does this PR do?

- Allows an Admin user edit and delete a security user.

- Add **put** and **delete** http verb to the existing _security-user_ endpoint.

- navigate to the **/api/v1/security-users** endpoint and try the edit and delete functionality.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -323,7 +323,17 @@ class SecurityUserViewSet(ModelViewSet):
     queryset = SecurityUser.objects.all()
     permission_classes = [IsAuthenticated, IsAdminUser]
     authentication_classes = [FirebaseTokenAuthentication, ]
-    http_method_names = ['get', 'post']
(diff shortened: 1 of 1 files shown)


api/views.py
```diff
@@ -323,7 +323,17 @@ class SecurityUserViewSet(ModelViewSet):
     queryset = SecurityUser.objects.all()
     permission_classes = [IsAuthenticated, IsAdminUser]
     authentication_classes = [FirebaseTokenAuthentication, ]
-    http_method_names = ['get', 'post']
+    http_method_names = ['get', 'post', 'put', 'delete']
+
+    def destroy(self, request, *args, **kwargs):
```
egichuri: is there a reason we need to override `destroy`?

precious-ijege: The default destroy method doesn't return a response, so I had to override it to return a message.

PR #203, opened 2018-11-14
[#161948300] set admin attribute in token

egichuri:


api/tests/test_user.py
```diff
@@ -211,16 +211,6 @@ def test_admin_user_get_users_from_api_endpoint(self, mock_verify_token):
         self.assertEqual(len(response.data['results']), User.objects.count())
         self.assertEqual(response.status_code, 200)
 
-    @patch('api.authentication.auth.verify_id_token')
-    def test_user_not_found_from_api_endpoint(self, mock_verify_token):
-        mock_verify_token.return_value = {'email': 'unavailable@email.com'}
-        response = client.get(
```
egichuri: We don't need this test. If a token is not valid `verify_id_token` will raise an error and no email will be returned

egichuri:
@esirK correct

PR #204, opened 2018-11-15
[#161948300] use post_save for firebase custom attributes

egichuri:
https://www.pivotaltracker.com/story/show/161948300


egichuri:
@esirK seems like using a post_save is better here to avoid having the FE do multiple requests if the attributes are not set

PR #205, opened 2018-11-15
[#160246264]: Sync users on AIS to ART (WIP)

nelmandela:
### What does this PR do?

    •Enables syncing of users on AIS to ART

    •Create scripts to retrieve updated users on AIS daily at 3.00 am

### Relevant pivotal tracker stories:

[#160246264](https://www.pivotaltracker.com/story/show/160246264)


scripts/sync_users.py
```diff
@@ -0,0 +1,60 @@
...
+import sys
+import django
+import re
+
+import requests
+from django.contrib.auth import get_user_model
+
+url = "https://api-prod.andela.com/api/v1/users"
+
+headers = {'Authorization': os.getenv('TOKEN')}
```
egichuri: Looks like [AIS](https://github.com/andela/engineering-playbook/wiki/Andela-API-HowTo#using-your-browser) uses `Bearer` not `TOKEN`

scripts/sync_users.py
```diff
@@ -0,0 +1,60 @@
...
+
+vlst = list(response.json().values())
+
+for value in vlst[0]: # NOQA
+
+    values = dict()
+    values.update({'email': value.get('email')})
+
+    if user.objects.filter(email=values.get('email')).exists():
+        continue
```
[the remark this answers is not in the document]
esirK: On the same, can we use the `get_or_create` method instead of the try-except ??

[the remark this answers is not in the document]
egichuri: yes.. that works. just confirm that it works with just the email provided.
`user, created = User.objects.get_or_create(email=...)`
then if `created` returns false you can check if you need to update the 3 fields

api/tests/test_user.py
```diff
@@ -97,20 +97,6 @@ def test_user_email_is_required(self):
                 cohort=20, slack_handle='@test_user1',
                 password='devpassword')
 
-    def test_user_cohort_is_required(self):
```
esirK: @egichuri According to [this](https://www.pivotaltracker.com/story/show/162193628) story, I think we need to maintain this test

egichuri: if so we can have a default cohort.. something like 00

egichuri: @MadgeKinyanjui

PR #208, opened 2018-11-19
[#162050415] Allows only a superuser to edit an asset location

esirK:
### What does this PR do?

    •Enables syncing of users on AIS to ART

    •Allows only a superuser to edit an asset location

    • Login as a normal user and try to update an asset location

### Relevant pivotal tracker stories:

[#162050415](https://www.pivotaltracker.com/story/show/162050415)

### DIFF
--- a/api/tests/test_manage_asset_api.py
+++ b/api/tests/test_manage_asset_api.py
@@ -444,3 +444,43 @@ class ManageAssetTestCase(APIBaseTestCase):
         response = client.get('{}/{}/'.format(self.manage_asset_urls, self.asset.uuid),
                               HTTP_AUTHORIZATION="Token {}".format(self.token_user))
         self.assertEqual(response.data.get('asset_location'), None)
+
+    @patch('api.authentication.auth.verify_id_token')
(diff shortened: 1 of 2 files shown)


api/tests/test_manage_asset_api.py
```diff
@@ -444,3 +444,19 @@ def test_admin_cannot_update_assets_to_no_existing_location(
...
+            "asset_code": "IC003",
+            "serial_number": "SN003",
+            "asset_location": "Nairobi",
+            "model_number": self.assetmodel.model_number}
+        AndelaCentre.objects.create(
+            centre_name="Nairobi",
+            country="Kenya")
+        res = client.put(
+            '{}/{}/'.format(self.manage_asset_urls, self.asset.uuid), data=data,
+            HTTP_AUTHORIZATION="Token {}".format(self.token_user))
```
egichuri: can we create an normal admin user for this test? (`is_staff=True` and is `is_superuser=False`)

esirK: This has been added.

PR #212, opened 2018-11-22
ft(Asset type has spec field): create a spec field

precious-ijege:
### What does this PR do?

- Creates a **has_spec** column in the **asset-types** model.

- Creates a new field in the model

- set a default value of **false**

- set admin right on the **asset-types** end point

- fix failing tests.

- As an Admin navigate to the /api/v1/asset-types endpoint.

### Screenshots

screen shot 2018-11-22 at 9 20 42 pm


core/models/asset.py
```diff
@@ -156,6 +156,7 @@ class AssetType(models.Model):
         AssetSubCategory,
         on_delete=models.PROTECT
     )
+    specs = models.NullBooleanField(blank=False, null=True)
```
egichuri: let's change this to `has_specs`

core/models/asset.py
```diff
@@ -149,7 +149,11 @@ class AssetType(models.Model):
     asset_type = models.CharField(unique=True, max_length=50)
     created_at = models.DateTimeField(auto_now_add=True, editable=False)
     last_modified = models.DateTimeField(auto_now=True, editable=False)
-    asset_sub_category = models.ForeignKey(AssetSubCategory, on_delete=models.PROTECT)
+    asset_sub_category = models.ForeignKey(
+        AssetSubCategory,
+        on_delete=models.PROTECT
+    )
+    has_specs = models.NullBooleanField(blank=False, null=True)
```
egichuri: sorry, let's have a `default=False` and do away with `core/migrations/populate_specs_field.py`

egichuri: also delete `core/migrations/0001_auto_20181130_1125.py` and `core/migrations/0033_auto_20181121_2302.py`, then makemigrations again to have just one file

precious-ijege: Done!

PR #225, opened 2018-12-17
[#162714408] enable django silk

egichuri:
- Testing out Django Silk

> From [here](https://github.com/jazzband/django-silk#silk)
**SECURITY NOTE**: Because Silk stores all HTTP requests into the database in plain text, it will store the request's sensitive information into the database in plain text (e.g. users' passwords!). This is a massive security concern. 

~For this reason we'll only enable it in dev envs for now.~ Using a custom version til the issue is fixed

- I've also locked all package versions in this PR


egichuri:
Yeah.. looks like it won't work for debug only due to the middleware placement. looking for a fix for the security notice

PR #228, opened 2019-01-07
fix deployment issues

egichuri:


Dockerfile
```diff
@@ -4,7 +4,7 @@ LABEL application="artbackend"
 
 
 ARG SECRET_KEY='secret key'
-ARG DJANGO_SETTINGS_MODULE='settings.prod'
+ARG DJANGO_SETTINGS_MODULE='settings.dev'
```
WinstonKamau: The change to `settings.dev` may make docker containers built for production use the dev settings.

PR #242, opened 2019-01-25
[#163048613]: View logs related to all assets

nelmandela:
### What does PR do?

	•Allows admin to view logs related to all assets

       •Add permission class
       •Modify AssetLogViewset
       •Add test

	•Log in as admin & navigate to `asset-log` endpoint 

### Related pivotal tracker stories

	•[#163048613](https://www.pivotaltracker.com/story/show/163048613)

### Screenshots:

screen shot 2019-01-25 at 13 00 56


api/views/assets.py
```diff
@@ -149,7 +149,7 @@ def get_queryset(self):
 class AssetLogViewSet(ModelViewSet):
     serializer_class = AssetLogSerializer
     queryset = models.AssetLog.objects.all()
-    permission_classes = [IsSecurityUser]
+    permission_classes = [IsLogUser]
```
egichuri: I think we can do `(IsSecurityUser|IsAdminUser, )` instead of writing a new permission class

nelmandela:
@egichuri i've implemented the change you recommended of which it was working with *djangorestframework = "==3.7.7"* so i upgraded it to *djangorestframework = "==3.9.1"*,this works but fails on circleci-i suspect its because the builds are using cached packages.is there a way of rerunning it without cache?

egichuri:
will take a look

PR #246, opened 2019-02-05
[#163726521] filter by serial numbers and partial matches

egichuri:
https://www.pivotaltracker.com/story/show/163726521


nelmandela:
LGTM

PR #247, opened 2019-02-06
[#163755285] Tests coverage diff

egichuri:
https://www.pivotaltracker.com/story/show/163755285


egichuri:
@MadgeKinyanjui this should fix your tests - after `pipenv uninstall --all` then `pipenv install -d`

PR #252, opened 2019-02-20
Bug file upload

precious-ijege:
### What does PR do?

- Fix bugs on file imports.

- Check for CSV files with missing required headings.

- Write test for to check error is handled.

- Navigate to the `file upload ` endpoint http://127.0.0.1:8000/api/v1/upload/ and test it out

### Screenshots:
screen shot 2019-02-20 at 11 20 54 pm


nelmandela:
LGTM

precious-ijege:
@egichuri I had to change a sample file in the test suite. The file did not contain all the required fields.

egichuri:
@precious-ijege i don't think we should change that file. we should allow the file to be processed if it has the minimum required fields - `Model Number` and either `Asset Code`  or `Serial No` (or both)

core/constants.py
```diff
@@ -48,3 +48,7 @@
 YEAR_CHOICES = []
 for year in range(2013, (datetime.now().year + 1)):
     YEAR_CHOICES.append((year, year))
+
+CSV_REQUIRED_HEADING = {'Make', 'Type', 'Asset Code', 'Category', 'Sub-Category'}
+CSV_REQUIRED_HEADING_MODEL = CSV_REQUIRED_HEADING.union({'Model Number'})
+CSV_REQUIRED_HEADING_SERIAL = CSV_REQUIRED_HEADING.union({'Serial No'})
```
egichuri: I'd suggest we also include constants for all fields we have for consistency.
eg
```
MAKE = 'Make'
TYPE = 'Type'... etc

OPTIONAL_IMPORT_FIELDS = ((MAKE, 'Make'), (TYPE, 'Type'), etc)
REQUIRED_IMPORT_FIELDS = ...
```
The required fields should be `Model Number`, and  `Asset Code` or `Serial Number`.
If you wish to follow what you have you can do this but have constants for the field names
```
CSV_OPTIONAL_HEADING = {'Make', 'Type', 'Asset Code', 'Category', 'Sub-Category'}
CSV_REQUIRED_HEADING = {'Model Number'}
CSV_REQUIRED_HEADING_ASSET_CODE = CSV_REQUIRED_HEADING.union({'Asset Code'})
CSV_REQUIRED_HEADING_SERIAL = CSV_REQUIRED_HEADING.union({'Serial Number'})
```
let's also change `HEADING` to `HEADERS`

api/tests/missing_headings.csv
```diff
@@ -0,0 +1,5 @@
+Asset Code,Serial No.,Notes
```
egichuri: why did this file change? it's named `missing_headings`

core/constants.py
```diff
@@ -48,3 +48,40 @@
...
+MODEL_NUMBER = 'Model Number'
+SERIAL_NUMBER = 'Serial No.'
+ASSIGNED_TO = 'Assigned To'
+STATUS = 'Status'
+ASSET_MEMORY = 'Memory'
+VERIFIED = 'Verified'
+NOTES = 'Notes'
+STORAGE = 'Storage'
+ASSET_PROCESSOR_TYPE = 'Processor Type'
+YOM = 'YOM'
```
egichuri: Let's use these constants [here](https://github.com/AndelaOSP/art-backend/blob/20087a548d89ac318fd6dc8b60738531aaf6c5df/core/assets_import_helper.py#L39) as well, and wherever else these field names are used

api/tests/missing_headings.csv
precious-ijege: It is a new file, for testing. It checks for missing headings, especially the most important one which is **Model number**. Remember, we need a **model number** and either **asset code** or serial number in a sample CSV file.

PR #257, opened 2019-02-27
Print asset details to file

precious-ijege:
### What does PR do?

- Exports assets details to a spreadsheet.

- create an **export-assets** endpoint.

- Allows an admin to print out assets details to a spreadsheet.

- Navigate to the **export-assets** endpoint _http://127.0.0.1:8000/api/v1/export-assets/_ and test it out

### Screenshots:
screen shot 2019-02-27 at 1 24 29 am


api/views/assets.py
```diff
@@ -444,3 +445,55 @@ def get(self, request):
...
+    authentication_classes = (FirebaseTokenAuthentication,)
+
+    def get(self, request):
+        assets = models.Asset.objects.all()
+        serializer = AssetSerializer(assets, many=True)
+        if len(serializer.data) == 0:
+            return Response({"error": "You have no assets"}, status=400)
+        self.create_sheet(serializer.data)
+        return Response({
+            "Success": "Aseest details have been exported successfully, Find file in root folder"}, status=200)
```
egichuri: ~Aseest~ typo

api/views/assets.py
```diff
@@ -444,3 +445,55 @@ def get(self, request):
...
         return response
+
+
+class ExportAssetsDetails(APIView):
+    permission_classes = [IsAuthenticated, IsAdminUser]
+    authentication_classes = (FirebaseTokenAuthentication,)
+
+    def get(self, request):
+        assets = models.Asset.objects.all()
+        serializer = AssetSerializer(assets, many=True)
```
egichuri: does this limit by asset location? we only want to export assets from the logged in users's location.

api/views/assets.py
precious-ijege: oops! I'll correct that.

api/views/assets.py
precious-ijege: It doesn't limit assets by location. From the story description on pivotal tracker, it states all assets.

egichuri: let's limit by the users location to avoid confusion when the user sees an asset on file that isn't on the app (since on the web app/API we filter by the users location).

precious-ijege: ok then. working on it.

api/views/assets.py
```diff
@@ -441,3 +442,55 @@ def get(self, request):
...
 
         return response
+
+
+class ExportAssetsDetails(APIView):
+    permission_classes = [IsAuthenticated, IsAdminUser]
+    authentication_classes = (FirebaseTokenAuthentication,)
+
+    def get(self, request):
+        assets = models.Asset.objects.filter(asset_location__name=request.user.location.name)
```
precious-ijege: It has been modified to filter assets with the users location.

api/views/assets.py
```diff
@@ -441,3 +442,55 @@ def get(self, request):
...
+        asset_types = []
+        for asset in assets_list:
+            if asset.get('asset_type') not in asset_types:
+                asset_types.append(asset.get('asset_type'))
+
+        workbook = xlsxwriter.Workbook('assets.xlsx')
+        bold = workbook.add_format({'bold': True, 'bg_color': 'silver'})
+        for asset_type in asset_types:
+            worksheet = workbook.add_worksheet(asset_type)
+            worksheet.write('A1', 'Make', bold)
```
egichuri: We should use the constants you added here too. You might need to merge this one first though.

Pipfile
```diff
@@ -23,6 +23,7 @@ gunicorn = "==19.9.0"
 gevent = "==1.3.7"
 pycountry = "==18.12.8"
 responses = "==0.10.5"
+xlsxwriter = "*"
```
egichuri: let's pin a version here

api/views/assets.py
```diff
@@ -434,10 +435,74 @@ def get(self, request):
...
+        serializer = AssetSerializer(assets, many=True)
+        if len(serializer.data) == 0:
+            return Response({"error": "You have no assets"}, status=400)
+        self.create_sheet(serializer.data)
+        return Response(
+            {
+                "Success": "Assets details have been exported successfully, Find file in root folder"
+            },
+            status=200,
+        )
```
egichuri: we should return the url since this message is to the FE

precious-ijege: @egichuri can you please explain what you mean? I dont understand what you mean returning url.

egichuri: A working url for the location of the export file since the user might want to download it. The same way we're doing it for the [import](https://github.com/AndelaOSP/art-backend/blob/20087a548d89ac318fd6dc8b60738531aaf6c5df/api/views/assets.py#L397)

PR #259, opened 2019-03-15
[#163690184]: An incident report can be marked as newly reported

nelmandela:
### What does PR do?

	•Allow admin to mark incident reports as newly reported

        •Add state transitions field 
	   •Add transition constants

	•Navigate to incidence reports endpoint

### Related pivotal tracker stories

	•[#163690184(https://www.pivotaltracker.com/story/show/163690184)

### Screenshots:

Screen Shot 2019-03-18 at 11 55 43


core/constants.py
```diff
@@ -48,3 +48,17 @@
 YEAR_CHOICES = []
 for year in range(2013, (datetime.now().year + 1)):
     YEAR_CHOICES.append((year, year))
+
+NEWLY_REPORTED = "Newly reported"
+INTERNAL_ASSESSMENT = "Internel assessment"
+REQUIRES_REPAIR = "Requires repair"
```
MadgeKinyanjui: Please view this story to see how state transitions should be implemented https://www.pivotaltracker.com/story/show/164099111

nelmandela:
alright let me refactor that

core/models/asset.py
```diff
@@ -558,3 +558,13 @@ def __str__(self):
 
     class Meta:
         ordering = ['-id']
+
+
+class StateTransitions(models.Model):
+    asset_incident_report = models.ForeignKey(
+        AssetIncidentReport, on_delete=models.PROTECT
+    )
+    transitions = models.CharField(max_length=50, default=constants.NEWLY_REPORTED)
```
egichuri: let's call this `state` and restrict the choices

requirements.txt
```diff
@@ -0,0 +1,54 @@
+-i https://pypi.org/simple
```
egichuri: please delete this since we have Pipfile

api/serializers/__init__.py
```diff
@@ -22,6 +22,7 @@
     AssetStatusSerializer,
     AssetSubCategorySerializer,
     AssetTypeSerializer,
+    StateTransitionsSerializer,
```
egichuri: let's change this to `StateTransitionsSerializer`
Ideally the model, view, serializer names should be singular.

api/views/assets.py
```diff
@@ -521,3 +522,10 @@ def get(self, request):
         response["Content-Disposition"] = 'attachment; filename="{}"'.format(filename)
 
         return response
+
+
+class StateTransitionsViewset(ModelViewSet):
```
egichuri: this too

core/constants.py
```diff
@@ -48,3 +48,5 @@
 YEAR_CHOICES = []
 for year in range(2013, (datetime.now().year + 1)):
     YEAR_CHOICES.append((year, year))
+
+NEWLY_REPORTED = "Newly reported"
```
egichuri: You can add the other states as well

core/models/asset.py
```diff
@@ -558,3 +558,13 @@ def __str__(self):
 
     class Meta:
         ordering = ['-id']
+
+
+class StateTransitions(models.Model):
```
egichuri: this too

core/constants.py
nelmandela: that was to come in the next task, or i do it now?

egichuri: If there's a task for it it's fine..

nelmandela:
for new reports or existing ones?from my understanding of that is like keeping a history of incident reports, of which a new incident report doesn't have a history

egichuri:
For new reports.. when a new incident report is submitted we should have an entry in the state transitions table

core/models/asset.py
```diff
@@ -552,9 +552,36 @@ class AssetIncidentReport(models.Model):
     witnesses = models.TextField(null=True, blank=True)
     police_abstract_obtained = models.CharField(max_length=255)
     submitted_by = models.ForeignKey('User', null=True, on_delete=models.PROTECT)
+    state = models.CharField(max_length=50)
```
egichuri: i think we can do without this field since the `StateTransition` model has a foreign key pointing to this table

egichuri:
One last thing.. i see a couple of migration files.. please delete them and regenerate the migrations so that we'll have only one migrations file for this task.

PR #262, opened 2019-03-21
[#159469906] refactor slack incidence report

egichuri:


egichuri:
@precious-ijege - trying this out to see if it'll speed up the slack command. lemme know what you think

precious-ijege:
@egichuri I see you reduced the number of interactions with the user. Good move! Looks good. We can test the branch locally to find out.

egichuri:
I tested it and it seems to work fine..

precious-ijege:
Ok! How about testing? Then we can merge it and try it out on production.

PR #265, opened 2019-03-29
Bug: Fix repeating asset log-#164726859

kmazi:
- Update test to check for log_type

 - Add test

#### What does this PR do?
Fixes a bug: Prevent checkin or checkout of an asset more than once

* Update existing test
* Write new test
* Update save() to prevent checking in or checking out an asset more than once.

Run:
1. `python manage.py test api.tests.test_asset_log.AssetLogModelTest.test_verify_double_checkin_for_asset`

2. `python manage.py test api.tests.test_asset_log.AssetLogModelTest.test_verify_double_checkout_for_asset` 

Formerly a security user could checkin an asset more than once or checkout an asset that has been checked out already.

https://www.pivotaltracker.com/story/show/164726859

### DIFF
--- a/api/tests/test_asset_log.py
+++ b/api/tests/test_asset_log.py
@@ -9,6 +9,7 @@ from rest_framework.test import APIClient
 # App Imports
 from api.tests import APIBaseTestCase
 from core.models import Asset, AssetLog, AssetModelNumber
+from core.constants import CHECKIN, CHECKOUT
 
 User = get_user_model()
 client = APIClient()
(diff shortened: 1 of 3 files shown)


api/tests/test_asset_log.py
```diff
@@ -43,6 +43,23 @@ def test_add_checkin(self):
             log_type="Checkin",
         )
         self.assertEqual(AssetLog.objects.count(), 3)
+        created_log = AssetLog.objects.filter(asset=self.test_other_asset).first()
+        self.assertEqual(created_log.log_type, "Checkin")
```
egichuri: could we use the constants defined [here](https://github.com/AndelaOSP/art-backend/blob/696da34d0de7ce234645e4df42899c9798352cf2/core/constants.py#L25) for "Checkin" and "Checkout"? we can also update the rest of the usage on this file.

core/models/asset.py
```diff
@@ -359,6 +359,9 @@ def clean(self):
 
     def save(self, *args, **kwargs):
         self.full_clean()
+        log = AssetLog.objects.filter(asset=self.asset).first()
+        if log and log.log_type == self.log_type:
+            return None
```
egichuri: let's raise an exception here instead of returning `None`. (`ValidationError` should do). Then we should also update the API so that it will be clear in the response that the asset has already been checked in or out

api/tests/test_asset_log.py
```diff
@@ -43,6 +43,23 @@ def test_add_checkin(self):
...
+            asset=self.test_other_asset,
+            log_type="Checkin",
+        )
+        # Second log
+        AssetLog.objects.create(
+            checked_by=self.security_user,
+            asset=self.test_other_asset,
+            log_type="Checkin",
+        )
+        self.assertEqual(AssetLog.objects.count(), 3)
```
egichuri: It's not clear why the logs count is 3. a better way is to check that the 2nd log errors out. (see comment [here](https://github.com/AndelaOSP/art-backend/pull/265/files#diff-bea1207c116ad66b7b398915447bd091R364))

api/tests/test_asset_log.py
```diff
@@ -52,6 +69,21 @@ def test_add_checkout(self):
...
+            asset=self.test_other_asset,
+            log_type="Checkout",
+        )
+        # Second log
+        AssetLog.objects.create(
+            checked_by=self.security_user,
+            asset=self.test_other_asset,
+            log_type="Checkout",
+        )
+        self.assertEqual(AssetLog.objects.count(), 3)
```
egichuri: same as above

core/models/asset.py
kmazi: I've returned None on the model level and raise a ValidationError at the serializer level

PR #275, opened 2019-04-05
Feature: Allow admin to edit users is_staff attribute

kmazi:
- Update an existing test

 - Write new tests to assert implementation

* Enable patch in the user endpoint
* Enable admin to change a user's is_staff attribute from api endpoint

Admin users could only change user's is_staff attribute from admin end

https://www.pivotaltracker.com/story/show/164238130


api/tests/test_user.py
```diff
@@ -238,14 +238,48 @@ def test_add_user_api_endpoint_cant_allow_put(self, mock_verify_token):
...
+        mock_verify_token.return_value = {'email': self.admin_user.email}
+        user = User.objects.first()
+        update_data = {'is_staff': True}
+        response = client.patch(
+            '{}/{}/'.format(self.users_url, user.id),
+            data=update_data,
+            HTTP_AUTHORIZATION="Token {}".format(self.token_admin),
+        )
+        self.assertEqual(response.status_code, 200)
+        is_staff = User.objects.first().is_staff
```
egichuri: would be safer to fetch user by id. please also add an assertion to check that other fields haven't changed.

egichuri (changes requested):

api/tests/test_user.py
kmazi: Adding check that other fields haven't changed would be me testing that the patch function works well; Django has test for that already. Patch updates only the attributes that are passed via request.

egichuri (dismissed):

kmazi:
@egichuri would need approval.

egichuri (approved):

PR #285, opened 2019-04-24
get assets assigned to a department

kenneth051:
#### What does this PR do?
get assets assigned to a department

We can assign assets to a department but we cannot get them. This PR makes it able for assets assigned to a department to be got.

1. On postman, goto url `http://127.0.0.1:8000/api/v1/departments/<department id>` and do a get request.

Ensure that you have assigned assets to that department first

https://www.pivotaltracker.com/n/projects/2146417/stories/165053155

Screen Shot 2019-04-26 at 12 38 51


api/serializers/andela_centres.py
```diff
@@ -39,8 +39,35 @@ class Meta:
...
+        fields = ("name", "id", "assets_assigned")
+
+    def get_assets_assigned(self, obj):
+        """This method returns assets assigned to a particluar department
+
+        Args:
+            obj (object): current object instance being fetched
+
+        Returns:
+            json : serailised assets belonging to the specified department
```
egichuri: typo - `serailised`

api/serializers/andela_centres.py
```diff
@@ -39,8 +39,35 @@ class Meta:
...
+        """
+
+        from api.serializers.assets import DepartmentAssetSerializer
+
+        department_assignee = models.AssetAssignee.objects.filter(
+            department_id=obj.id
+        ).first()
+        assets = models.Asset.objects.filter(assigned_to=department_assignee)
+        serialized_assets = DepartmentAssetSerializer(assets, many=True)
```
egichuri, line 65: why not use the `AssetSerializer`?

api/serializers/assets.py
```diff
@@ -437,6 +437,12 @@ class Meta:
         fields = ("asset_type", "model_number", "count_by_status")
 
 
+class DepartmentAssetSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.Asset
+        fields = ("id", "uuid", "asset_category")
```
egichuri: If we are to retain this and not use the `AssetSerializer` we should include the serial number and asset code (and probably the asset type). also exclude the id since we fetch assets by uuid.

api/serializers/andela_centres.py
kenneth051: The asset serializer calls the department serializer. if I call the asset serializer, it would create a recursive chain.

api/serializers/assets.py
kenneth051: I have included those feilds

api/serializers/andela_centres.py
kenneth051: it is serialized now

egichuri (approved):

kmazi (approved):

PR #290, opened 2019-05-02
#164099111 Allow an admin to update an incident report's state.

RonKbS:
## What does this PR do?
- This PR enables an admin to add `requires repair`, `requires external assessment`,  `damaged`  to an incident report concerning an asset which will be saved in the state-transition table.

When a report is newly reported, it is marked newly_reported in the database. This PR allows an Admin to update the report and asset states to the choices described above.

1. On the browser, go to URL `http://127.0.0.1:8000/api/v1/state-transitions/{id}` and update an incident report's state using the format below;

```
{
            "incident_report_state": "internal assessment",
            "asset_state": "damaged",
}
```

A response with the updated information will be returned.
Note: To obtain the status ids, the same URL can be used in conjunction with the get method on a client like Postman or Insomnia.

[#164099111](https://www.pivotaltracker.com/story/show/164099111)


RonKbS:
> #### What does this PR do?
> * This PR enables an admin to add requires repair`, `requires external assessment`,  `damaged`or`Lost` to an incident report concerning a utility which will be saved in the transition table.
> 
> #### Description of Task to be completed?
> When a report is newly reported, it is marked newly_reported in the database. However, no detail is provided as to what is wrong with the item that has been mentioned in the report. This PR enables an admin to add this detail
> 
> #### How should this be manually tested?
> 1. On browser, goto url `http://127.0.0.1:8000/api/v1/incidence-reports/` and post an incident report to which an addition field (state_transition_detail) is added. A 201 response should be received.
> 
> #### What are the relevant pivotal tracker stories?
> [#164099111](https://www.pivotaltracker.com/story/show/164099111)

api/views/assets.py
```diff
@@ -279,6 +279,20 @@ def get_queryset(self):
...
         serializer.save(submitted_by=self.request.user)
 
+        # update the state_transition with option from admin
+        state_transition_object = models.StateTransition.objects.filter(
+            asset_incident_report_id=models.AssetIncidentReport.objects.filter(
+                asset_id=serializer.context['request'].data.get('asset', None)
+            )
+            .first()
+            .id
+        ).first()
```
egichuri: could we simplify this to make it more readable?
```
asset_id = ...
incidence_report = ...
state_transition_object = ...
```

core/models/asset.py
```diff
@@ -586,6 +586,11 @@ class StateTransition(models.Model):
         'AssetIncidentReport', on_delete=models.PROTECT
     )
     state = models.CharField(max_length=50, default=constants.NEWLY_REPORTED)
+    state_transition = models.CharField(
```
egichuri: I believe this is covered by the state field.
`Add internal assessment option in transition state field . with options requires repair , requires external assessment, damaged or lost`

egichuri: hmm.. seeing how this might be confusing - the idea is when an incidence report is submitted the admins can transition it from one state to the next.
- newly reported
- internal assessment
  - requires repair
  - requires external assessment
  - damaged
  - lost

i suggest we add a field that identifies whether a state has a 'parent' or not

egichuri (changes requested):

core/models/asset.py
RonKbS: As I understand it, the additional field added is to provide more detail (lost, damaged etc) on submitting a report.
Are you suggesting to add another aside from that? If so, kindly elaborate on 'parent'.

egichuri: we always start with a `newly reported` state. after this the asset undergoes internal assessment the department decides whether it requires external assessment, repair or should be marked as damaged.
in this case the states would be something like:
`newly reported`=> `internal assessment` => `external assessment`=> `out for repair`

the options `requires repair`, `requires external assessment`, `damaged` belong to the `internal assessment` state. we can go with details but the field name should match the functionality.

please discuss this further with @MadgeKinyanjui in case i've missed something

api/tests/test_asset_incident_report_api.py
```diff
@@ -32,6 +32,7 @@ def test_authenticated_user_post_incident_report(self, mock_verify_id_token):
             "loss_of_property": "Mobile Phone",
             "witnesses": "John Doe +2347548458457",
             "police_abstract_obtained": "Yes",
+            "state_transition_detail": "lost",
```
kmazi: Your assertion should show that the state_transition_detail was added to the database after `Post`.

kmazi (changes requested):

api/tests/test_asset_incident_report_api.py
```diff
@@ -141,3 +141,42 @@ def test_cant_allow_delete_incident_report(self, mock_verify_id_token):
...
+    @patch('api.authentication.auth.verify_id_token')
+    def test_require_both_fields_to_patch_incident_report_status(
+        self, mock_verify_id_token
+    ):
+        mock_verify_id_token.return_value = {'email': self.admin_user.email}
+        data = {
+            "asset_incident_report": self.report_status.asset_incident_report.id,
+            "incident_report_state": "internal assessment",
+        }
+        response = client.put(
```
kmazi: You're testing that you could `patch` an incident report status but you're making a `put` request here.

api/tests/test_asset_incident_report_api.py
```diff
@@ -141,3 +141,42 @@ def test_cant_allow_delete_incident_report(self, mock_verify_id_token):
...
+
+    @patch('api.authentication.auth.verify_id_token')
+    def test_can_allow_patch_incident_report_status(self, mock_verify_id_token):
+        mock_verify_id_token.return_value = {'email': self.admin_user.email}
+        data = {
+            "asset_incident_report": self.report_status.asset_incident_report.id,
+            "incident_report_state": "internal assessment",
+            "asset_state": "damaged",
+        }
+        response = client.put(
```
kmazi: You test method says you could `patch` a report but you're making a `put` request here.

kmazi (changes requested):

core/models/asset.py
```diff
@@ -576,16 +576,43 @@ def _save_initial_incident_report_state(self):
...
+    )
 
     class Meta:
         verbose_name_plural = 'State Transitions'
+
+    def save(self, *args, **kwargs):
+
+        # admin updates asset_status if damaged
+        if self.asset_state == "damaged":
+            asset_status = AssetStatus.objects.filter(
```
egichuri: we should be creating a new asset status here instead of updating the existing one. we normally do this for historical data.

core/models/asset.py
```diff
@@ -576,16 +576,43 @@ def _save_initial_incident_report_state(self):
...
     asset_incident_report = models.ForeignKey(
         'AssetIncidentReport', on_delete=models.PROTECT
     )
-    state = models.CharField(max_length=50, default=constants.NEWLY_REPORTED)
+    incident_report_state = models.CharField(
+        max_length=50,
+        choices=constants.REPORT_STATE_OPTIONS,
+        default=constants.NEWLY_REPORTED,
+    )
+    asset_state = models.CharField(
```
egichuri: we can work with the AssetStatus model we currently have

core/models/asset.py
RonKbS: The implementation of the AssetStatus model is such that the previous status is saved once this code is hit. So it performs the update but keeps the previous status as well as can be shown in the screenshot.
Screenshot 2019-05-09 at 14 41 25

core/models/asset.py
RonKbS: If I'm to recall the sync I had with Madge, we are to only update the asset's state (in the AssetStatus model), if damaged is passed. This leaves the other options for the asset (requires repair and requires external assessment) with a need of a field to be placed in. And so this additional field was created. A note was also added in the PT story regarding this.

core/models/asset.py
egichuri: this actually happens when a new status is added. if you add a `damaged` state the previous state will also be saved.

egichuri: see [this](https://github.com/AndelaOSP/art-backend/blob/ab3a9ed77f0a92cdc9cf870956bce193dd28e768/core/models/asset.py#L396)

core/models/asset.py
egichuri: oh.. cool. could we give it a name that won't be confused with the [asset status](https://github.com/AndelaOSP/art-backend/blob/ab3a9ed77f0a92cdc9cf870956bce193dd28e768/core/constants.py#L18)?

core/models/asset.py
RonKbS: Yes, so I am not sure if I am understanding your point. There is a new status being created here. It replaces the old one which is saved as previous_status.

egichuri: we should just do `AssetStatus.objects.create(current_status=constants.DAMAGED, asset_id=self.asset_incident_report.asset_id)` instead of filtering the last added status

core/models/asset.py
RonKbS: I have updated it to `asset_state_from_report`.

egichuri (approved):

api/tests/test_asset_incident_report_api.py
RonKbS: I've updated the tests to work with client.patch, @kmazi.

kmazi (approved):

PR #291, opened 2019-05-07
Add serial number and asset code filter to assetlog query endpoint

kmazi:
## What does this PR do?
Enable asset-log filtering by asset code and asset serial number

- Update tests for checks that asset logs could be filtered by serial number and asset code.
- Update the asset-log filter to include asset-serial-number and asset-code fields.

1. Query for assetlogs using its endpoint (api/v1/asset-logs/?asset_serial=`serial_value`&asset_code=`asset_code_value`\)

https://www.pivotaltracker.com/story/show/165785995

### DIFF
--- a/api/tests/test_asset_log.py
+++ b/api/tests/test_asset_log.py
@@ -230,6 +230,78 @@ class AssetLogModelTest(APIBaseTestCase):
         )
 
     @patch("api.authentication.auth.verify_id_token")
+    def test_authenticated_admin_user_get_asset_logs_filtered_by_serial_number(
+        self, mock_verify_id_token
+    ):
+        mock_verify_id_token.return_value = {"email": self.admin_user.email}
+        AssetLog.objects.create(
+            checked_by=self.security_user, asset=self.test_other_asset, log_type=CHECKIN
+        )
+        asset_logs_url = (
+            f"{self.asset_logs_url}/?asset_serial={self.test_other_asset.serial_number}"
+        )
+        response = client.get(
+            asset_logs_url, HTTP_AUTHORIZATION=f"Token {self.token_admin}"
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.data["results"]), 1)
+        self.assertEqual(
+            response.data["results"][0]["asset"],
+            f"{self.test_other_asset.serial_number} - {self.test_other_asset.asset_code}",
+        )
+
+    @patch("api.authentication.auth.verify_id_token")
+    def test_admin_user_get_asset_logs_filtered_by_invalid_serial_number(
+        self, mock_verify_id_token
+    ):
+        mock_verify_id_token.return_value = {"email": self.admin_user.email}
+        AssetLog.objects.create(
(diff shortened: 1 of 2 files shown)


api/tests/test_asset_log.py
```diff
@@ -229,6 +229,78 @@ def test_authenticated_admin_user_get_filtered_list_of_asset_logs(
...
+            checked_by=self.security_user, asset=self.test_other_asset, log_type=CHECKIN
+        )
+        asset_logs_url = (
+            f"{self.asset_logs_url}/?asset_serial={self.test_other_asset.serial_number}"
+        )
+        response = client.get(
+            asset_logs_url, HTTP_AUTHORIZATION=f"Token {self.token_admin}"
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.data["results"]), 1)
```
egichuri, line 247: checking the response length can be flaky - if a new asset log is created in the base class setup this would break. could we check that all items in the result contain the serial number the filter was for?

api/tests/test_asset_log.py
```diff
@@ -229,6 +229,78 @@ def test_authenticated_admin_user_get_filtered_list_of_asset_logs(
...
+
+    @patch("api.authentication.auth.verify_id_token")
+    def test_admin_user_get_asset_logs_filtered_by_invalid_serial_number(
+        self, mock_verify_id_token
+    ):
+        mock_verify_id_token.return_value = {"email": self.admin_user.email}
+        AssetLog.objects.create(
+            checked_by=self.security_user, asset=self.test_other_asset, log_type=CHECKIN
+        )
+        asset_logs_url = f"{self.asset_logs_url}/?asset_serial=QWS^&112"
```
egichuri: `&` in a query param value could have an undesirable effect

egichuri (changes requested):

api/tests/test_asset_log.py
kmazi: If a new asset-log is created in the base class setup, many tests would break. Tests are suppose to be stand alone. Test objects created in the base class are to be shared by all or most test methods. It is for this reason I've created an asset-log in the test method just for this assertion and thus checking for the response length.

kmazi: I think checking that all items in the result contain the serial number won't scale because the returned asset object in the log is just a string representation and wasn't serialized.

egichuri: that's what you're doing [here](https://github.com/AndelaOSP/art-backend/pull/291/files#diff-1b7c94b7be9d3e42c12b30660d17e1f9R250) though - for one object.

kmazi: Yeah. That's because it's the object I created in the test method; it's just ok to have that [here](https://github.com/AndelaOSP/art-backend/pull/291/files#diff-1b7c94b7be9d3e42c12b30660d17e1f9R250). If the response were to be a list of logs, checking that items in the result contain the serial number would also match for substrings which shouldn't be the case.

kmazi: @egichuri

egichuri (approved):

PR #295, opened 2019-05-10
[#165373121] Allow user to upload police abstract

nelmandela:
#### What does this PR do?

  •Allow user to upload police abstract

  •Given an andelan with asset allocated
  •When I want to submit an incident report on slack
  •Then on the police abstract bit, I can attach an image or upload a pdf.

 •On browser, goto `http://127.0.0.1:8000/api/v1/incidence-reports` and file an incidence report 

  [#165373121](https://www.pivotaltracker.com/story/show/165373121)

#### Screenshots 
Screen Shot 2019-05-13 at 10 54 27


api/views/assets.py
```diff
@@ -594,3 +596,22 @@ def perform_update(self, serializer):
                     )
                 }
             )
+
+
+class PoliceAbstract(APIView):
```
RonKbS: Kindly share your reason for adding this view seeing as I do not see it being used.

core/models/asset.py
```diff
@@ -19,6 +19,10 @@
 logger = logging.getLogger(__name__)
 
 
+def user_abstract(user, filename):
+    return f'user_{user.id}_{filename}'
```
RonKbS: I'm not sure this function is returning the right thing when you call it while uploading the file in the incident_report model. 
The file I uploaded for example is labelled `user_None_WASH_OFFICER_ADVERT.pdf`. The original file name is `WASH OFFICER ADVERT.pdf`.
The object `user` argument is capturing the `AssetIncidentReport` object. I would recommend that you reimplement this to return a proper name.

RonKbS (commented):
I've been able to upload a file using postman as shown below. Great work!!
Screenshot 2019-05-14 at 16 52 37

I have a few recommendations that I request you look at.

api/views/assets.py
nelmandela: its for authentication purposes, linking the uploaded file to logged in user and implementing file storage

core/models/asset.py
nelmandela: on it

nelmandela: @RonKbS 
mind giving it a look

RonKbS: It's now returning a `user_email_file_name` format 👍🏽

RonKbS:
To cover the changes you've introduced, I recommend that you look up ways to mock file uploading. Here, for example, => `https://stackoverflow.com/a/27345260/9904689`

kmazi:
@nelmandela Is there a blocker preventing the completion of this story?

nelmandela:
@kmazi yes...i'm having issues with slack configuration that i'm working on

api/views/assets.py
```diff
@@ -298,7 +300,17 @@ def get_queryset(self):
...
-        serializer.save(submitted_by=self.request.user)
+        abstract = self.request.FILES.get('police_abstract', None)
+        user = self.request.user
+        if abstract:
+            abstract_name = user_abstract(user, abstract.name)
+            serializer.save(submitted_by=user, police_abstract=abstract_name)
+            fs = FileSystemStorage()
+            filename = fs.save(abstract_name, abstract)
+            uploaded_file_url = fs.url(filename)
+            return Response(data={"success": {"uploaded_file_url": uploaded_file_url}})
```
kmazi: The response seems to not be consistent when police abstract is uploaded and when it's not uploaded. This means when uploaded, the response contains only the uploaded_file_url and when not uploaded, it contains the submitted incident_report fields. I think it would be more appropriate if the uploaded_file_url is the additional field to the response when police abstract is uploaded.

api/serializers/assets.py
```diff
@@ -471,18 +471,7 @@ class AssetIncidentReportSerializer(serializers.ModelSerializer):
...
-            "incident_type",
-            "incident_location",
-            "incident_description",
-            "injuries_sustained",
-            "loss_of_property",
-            "witnesses",
-            "submitted_by",
-            "police_abstract_obtained",
-        )
+        fields = '__all__'
```
kmazi: This isn't a secured way to expose all fields. It's more appropriate to list out the fields to serialize and de-serialize explicitly to avoid accidentally exposing a field which should be hidden.

kmazi (changes requested):

kmazi:
There isn't any integration test for this feature also.

kmazi (approved):

PR #296, opened 2019-05-14
#164101156 Allow an admin to update an incident report's state under external assessment

RonKbS:
## What does this PR do?
- This PR ensures an admin to mark an asset's status in an incident report (under external assessment) as either `Damaged` or `requires repair`.

Currently, an incident report marked under external assessment can have the asset status also updated to `requires external assessment`. This PR restricts this, therefore, allowing an admin to only mark an asset's status in a report as `Damaged` or `requires repair`.

1. On the browser, go to URL `http://127.0.0.1:8000/api/v1/state-transitions/{id}` and update an incident report's state using the format below;

```
{
            "incident_report_state": "external assessment",
            "asset_state": "Damaged",
}
```

A response with the updated information will be returned.
Note: To obtain the status ids, the same URL can be used in conjunction with the get method on a client like Postman or Insomnia.

[#164101156](https://www.pivotaltracker.com/story/show/164101156)

### DIFF
--- a/api/views/assets.py
+++ b/api/views/assets.py
@@ -576,6 +576,17 @@ class StateTransitionViewset(ModelViewSet):
     def perform_update(self, serializer):
 
         try:
+            report_sate = self.request.data.get("incident_report_state", None)
+            asset_state = self.request.data.get("asset_state_from_report", None)
+            if (
+                report_sate == "external assessment"
+                and asset_state == "requires external assessment"
+            ):
+                raise serializers.ValidationError(
+                    {
+                        "Error": "Asset state option is not valid for given report state"
+                    }
+                )
             serializer.save(
                 incident_report_state=self.request.data.get(
                     "incident_report_state", None
--- a/api/views/assets.py
+++ b/api/views/assets.py
@@ -583,9 +583,7 @@ class StateTransitionViewset(ModelViewSet):
                 and asset_state == "requires external assessment"
             ):
                 raise serializers.ValidationError(
-                    {
-                        "Error": "Asset state option is not valid for given report state"
-                    }
+                    {"Error": "Asset state option is not valid for given report state"}
                 )
             serializer.save(
                 incident_report_state=self.request.data.get(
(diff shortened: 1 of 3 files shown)


egichuri (approved):

api/views/assets.py
```diff
@@ -576,6 +576,15 @@ class StateTransitionViewset(ModelViewSet):
     def perform_update(self, serializer):
 
         try:
+            report_sate = self.request.data.get("incident_report_state", None)
+            asset_state = self.request.data.get("asset_state_from_report", None)
+            if (
+                report_sate == "external assessment"
+                and asset_state == "requires external assessment"
+            ):
+                raise serializers.ValidationError(
```
kmazi: I don't think it's nice raising exceptions in a try block which is meant to catch exceptions. You could just do your data integrity check and raise this exception outside the try block.

api/views/assets.py
```diff
@@ -576,6 +576,15 @@ class StateTransitionViewset(ModelViewSet):
     def perform_update(self, serializer):
 
         try:
+            report_sate = self.request.data.get("incident_report_state", None)
```
kmazi: I'm sure you meant to write `report_state`. loooooool

api/views/assets.py
```diff
@@ -576,6 +576,15 @@ class StateTransitionViewset(ModelViewSet):
...
+            if (
+                report_sate == "external assessment"
+                and asset_state == "requires external assessment"
+            ):
+                raise serializers.ValidationError(
+                    {"Error": "Asset state option is not valid for given report state"}
+                )
             serializer.save(
                 incident_report_state=self.request.data.get(
                     "incident_report_state", None
```
kmazi: Since you created the `asset_state` and `report_sate` variables, you could use them here rather that reading off request again. You could do the same for the one below.

kmazi (changes requested):

kmazi (approved):

PR #301, opened 2019-05-22
#165376778 Add fields to user and asset models.

RonKbS:
## What does this PR do?
This PR adds department fields to user and asset tables. It also adds active and paid_postpaid fields to the asset table as well. A table is also added to track owners of assets, different from people to whom assets are assigned. Third, the PR restricts users to only view assets in their departments and location

Currently, tracking of which departments users and assets belong to is not being done. There is also no tracking of actual owners (department, user or workspace) of an asset but only an assignee. Thirdly, users can view assets irrespective of which departments or locations they belong. The changes in this PR address these issues.

1. On the browser, go to URL `http://127.0.0.1:8000/api/v1/users/` and add a user with an additional key-value for `department` for an existent department. A response with the given department should be returned

2. Go to `http://127.0.0.1:8000/api/v1/assets/` and post an asset with additional key-value pairs for `department`, `active` and `prepaid_postpaid`. A response with the same data should be returned.

3. If there are any existing assets not in the same department or location as a user interacting with `http://127.0.0.1:8000/api/v1/assets/` using a get method, these assets will not appear in the response.

4. Go to `http://127.0.0.1:8000/api/v1/manage-assets/{uuid}/` and patch an asset's active, prepaid_postpaid or department fields. A response with the updated asset field should be returned. This is dependent on which type of asset is being worked on, with only simcards and mifis allowed to have their active and prepaid_postpaid fields.

5. Endpoints that were previously using current_owner/previous_owner are now using current_assignee/previous_assignee.

[#165376778](https://www.pivotaltracker.com/story/show/165376778)


api/views/assets.py
```diff
@@ -111,6 +111,10 @@ class AssetViewSet(ModelViewSet):
     def get_queryset(self):
         user = self.request.user
         query_filter = {}
+
+        # view only assets in ones location and ones department
+        query_filter = {"department": user.department, "asset_location": user.location}
+
```
egichuri: this should be on `ManageAssetViewSet`

core/constants.py
```diff
@@ -111,3 +111,8 @@
     (REQUIRES_EXTERNAL_ASSESSMENT, "requires external assessment"),
     (DAMAGED, "Damaged"),
 )
+
+PAID = "paid"
```
egichuri: prepaid

api/views/assets.py
```diff
@@ -111,6 +111,10 @@ class AssetViewSet(ModelViewSet):
     def get_queryset(self):
         user = self.request.user
         query_filter = {}
+
+        # view only assets in ones location and ones department
+        query_filter = {"department": user.department, "asset_location": user.location}
```
egichuri: this should be on `ManageAssetViewSet`

egichuri: for clarity - `ManageAssetViewSet` is for admin users - the one we use for the FE dashboard, `AssetViewSet` is for security users (for checkin/checkout) and andela users (to access assets assigned to them.)

egichuri (changes requested):

api/views/assets.py
RonKbS: The PT story does not specify which user_type is to be restricted to seeing assets for their teams and locations. So to confirm, admin users will be the ones with this restriction?

egichuri: yeah.. only admin users.

kmazi:
@egichuri How about this one, should I go on to merge?

egichuri (approved):

api/serializers/assets.py
```diff
@@ -128,6 +140,43 @@ def to_internal_value(self, data):
...
+        asset_type = instance.model_number.asset_make.asset_type.name
+        prepaid_or_postpaid = validated_data.get('prepaid_or_postpaid')
+        active_inactive = validated_data.get('active')
+        expiry_date = validated_data.get('expiry_date')
+        instance_type_and_associated_error = {
+            "simcard": {
+                "prepaid_or_postpaid": "Only sim cards can be prepaid or postpaid"
+            },
+            "mifi": {"active": "Only mifi cards can be activated or deactivated"},
+            "atmcard": {"expiry_date": "Only atm cards can have this field updated"},
```
MadgeKinyanjui: @RonKbS change this to emburse card and should add the field for active or inactive

api/serializers/assets.py
```diff
@@ -128,6 +140,43 @@ def to_internal_value(self, data):
...
+    # # allow updating of active_inactive and paid_prepaid for specific asset_types
+    def update(self, instance, validated_data):
+
+        asset_type = instance.model_number.asset_make.asset_type.name
+        prepaid_or_postpaid = validated_data.get('prepaid_or_postpaid')
+        active_inactive = validated_data.get('active')
+        expiry_date = validated_data.get('expiry_date')
+        instance_type_and_associated_error = {
+            "simcard": {
+                "prepaid_or_postpaid": "Only sim cards can be prepaid or postpaid"
```
MadgeKinyanjui: Sim cards in place of active or inactive have a filed called 'plan' that has three options basic plan, business plan and none.

RonKbS:
@MadgeKinyanjui, the asked for changes have been implemented.

PR #302, opened 2019-05-30
user emails should be andela emails

kenneth051:
#### What does this PR do?
user emails should be Andela emails

Enforces only Andela emails to be used in the application

Try creating a superuser or normal user without giving that user an andela email, This will through a validation error.

https://www.pivotaltracker.com/n/projects/2146417/stories/164753083

### DIFF
--- a/core/models/user.py
+++ b/core/models/user.py
@@ -39,7 +39,6 @@ class UserManager(BaseUserManager):
     def create_superuser(self, **fields):
         fields.setdefault("is_staff", True)
         fields.setdefault("is_superuser", True)
-
         if fields.get("is_staff") is not True:
             raise ValueError("Superuser must have is_staff=True.")
         if fields.get("is_superuser") is not True:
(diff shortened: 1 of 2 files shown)


core/models/user.py
```diff
@@ -77,6 +76,7 @@ def clean(self):
             raise ValidationError("Only andela email addresses allowed")
 
     def save(self, *args, **kwargs):
+        self.clean()
```
kmazi, line 79: Is there any real reason to call `self.clean()` explicitly here seeing that `self.full_clean()` calls it internally?

kmazi (changes requested):

core/models/user.py
kenneth051: @kmazi  self.clean() was not being called initially, because if it were, then we would always get an error when creating a user without an Andela email. So I had to call it directly to make it work.

kmazi: Try putting a debugger when not calling it explicitly to see that it's called when `full_clean` is called.

kmazi (approved):

PR #303, opened 2019-06-06
#166114088 Assigner history

3N61N33R:
#### What does this PR do?
- adds assigner field to allocation history model

Given an admin with access
When I request to view an asset
Then I should see the list of assignee history with timestamps and the list of people who assigned it in different instances

Currently, we only have a list of the people assigned in the allocation history.

[#166114088](https://www.pivotaltracker.com/story/show/166114088)


kmazi (changes requested):
It's nice to have integration tests at least for your feature.

api/tests/test_allocation_api.py
```diff
@@ -24,16 +24,24 @@ def test_non_authenticated_user_view_assets(self):
     @patch("api.authentication.auth.verify_id_token")
     def test_get_allocations(self, mock_verify_id_token):
         """Test post new allocation"""
+
         mock_verify_id_token.return_value = {"email": self.other_user.email}
-        response = client.get(
+
+        data = {"asset": self.asset.id, "current_owner": self.asset_assignee.id}
+        client.post(
```
kmazi: Is this test for posting allocations or getting allocations? I also see that the method name says get but the description says post and also the line above.
For test purpose you could just create an additional AllocationHistory object and save it in the database.

api/tests/test_allocation_api.py
```diff
@@ -24,16 +24,24 @@ def test_non_authenticated_user_view_assets(self):
...
-            len(response.data["results"]), AllocationHistory.objects.count()
+        response = client.get(
+            self.allocations_urls, HTTP_AUTHORIZATION=f"Token {self.token_other_user}"
         )
+
+        data = response.data["results"]
+        self.assertEqual(len(data), AllocationHistory.objects.count())
         self.assertEqual(response.status_code, 200)
+        self.assertEqual(data[0]["assigner"], self.other_user.email)
+        self.assertIn("assigner", data[0])
```
kmazi: You already used the "assigner" key in the previous assertion which made this somewhat unnecessary.

kmazi (commented):
There seems to be no test that shows, as contained in the story description, that when a user views an asset then they should see a list of allocation history of both assigner and assignee.

3N61N33R:
> There seems to be no test that shows, as contained in the story description, that when a user views an asset then they should see a list of allocation history of both assigner and assignee.

This had already been tested for in a different PR on the assignee that had been merged
What is on this PR are the adjustments for the assigner

api/tests/test_allocation_api.py
3N61N33R: As written it is for `getting allocations` 
I added the post since the response had no allocations whatsoever and I couldn't assert that the response had the assigner
Hadn't looked at the docstring since I was adjusting the test, an oversight on my end
Will correct that, thanks

kmazi: You can create the allocation history here with AllocationHistory.objects.create() rather than making a post.

kmazi:
> > There seems to be no test that shows, as contained in the story description, that when a user views an asset then they should see a list of allocation history of both assigner and assignee.
> 
> This had already been tested for in a different PR on the assignee that had been merged
> What is on this PR are the adjustments for the assigner

Then let the tests show the adjustment you've made in the response when viewing details of an asset. Isn't that what's in the story? ```When I request to view an asset
Then I should see the list of assignee history with timestamps and the list of people who assigned it in different instances```

kmazi:
Your tests are more like a documentation of your implementation you know.

kmazi (approved):

PR #304, opened 2019-06-07
#165377092 Allow posting and getting of asset invoice receipts

RonKbS:
#### What does this PR do?
This PR enables an admin to view invoice_receipts of assets and gives a super_admin the ability to add an invoice_receipt to an asset

Currently, there is no way of posting an asset along with its invoice_receipt. This PR, enables this.

1. On the browser, go to URL http://127.0.0.1:8000/api/v1/manage-assets/ and add an asset along with a pdf file in the invoice_receipt field. A 201 response should be returned. This field can also be updated as long as the user is super_admin.
2. A normal admin can view a link to an invoice_receipt when using the GET method to view assets.

https://www.pivotaltracker.com/story/show/165377092


api/tests/test_manage_asset_api.py
```diff
@@ -102,11 +103,15 @@ def test_authenticated_admin_view_all_assets(self, mock_verify_id_token):
...
         mock_verify_id_token.return_value = {"email": self.admin_user.email}
+        receipt = SimpleUploadedFile(
+            "file.pdf", b"file_content", content_type="document/pdf"
+        )
         data = {
             "asset_code": "IC002",
             "serial_number": "SN002",
             "model_number": self.assetmodel.name,
             "purchase_date": "2018-07-10",
+            "invoice_receipt": receipt,
```
kmazi, line 115: Would the invoice_receipt be in the response after the post request? If it would then be good to assert for it in the tests.

kmazi: You can also test that when an admin fetches an asset, then they also fetch the asset invoice_receipt.

api/views/assets.py
```diff
@@ -78,6 +78,11 @@ def get_object(self):
         obj = get_object_or_404(queryset, uuid=self.kwargs["pk"])
         return obj
 
+    @staticmethod
+    def restrict_to_only_super_users(invoice_receipt, user):
+        if invoice_receipt and user.is_superuser is not True:
```
kmazi: `user.is_superuser` is a boolean value so you could use it directly here:`if invoice_receipt and not user.is_superuser`

api/views/assets.py
```diff
@@ -86,9 +91,20 @@ def create(self, request, *args, **kwargs):
         return response
 
     def perform_create(self, serializer):
+
+        ManageAssetViewSet.restrict_to_only_super_users(
```
kmazi: Also seeing that aside checking if the user is a superuser it also checks for the presence of an invoice_receipt, making this a static method means this could be used outside here and that is unlikely. You could make the `restrict_to_only_super_users` method an internal method if you want.

api/tests/test_manage_asset_api.py
```diff
@@ -121,6 +126,24 @@ def test_admin_can_post_asset(self, mock_verify_id_token):
         self.assertEqual(Asset.objects.count(), count + 1)
         self.assertEqual(response.status_code, 201)
 
+    @patch("api.authentication.auth.verify_id_token")
+    def test_non_super_admin_cannot_patch_invoice_receipt(self, mock_verify_id_token):
```
kmazi, line 153: Can a super_admin patch invoice receipt?

kmazi (changes requested):

api/tests/test_manage_asset_api.py
RonKbS: Yes, a super_admin can patch an invoice receipt.

kmazi: Could you add test to show that super admin can patch invoice receipt if you've not already done that.

kmazi (approved):

PR #308, opened 2019-06-20
show categorization on asset logs response

kenneth051:
[Finishes #166728438]

#### What does this PR do?
show categorization on asset logs response

an asset log will show details about the asset it belongs to like asset category, asset make, asset type and asset subcategory type

n postman, go to http://127.0.0.1:8000/api/v1/asset-logs You will see more asset details of that specific asset log.

N/A

https://www.pivotaltracker.com/n/projects/2146417/stories/166728438


api/tests/test_asset_log.py
```diff
@@ -784,3 +784,27 @@ def test_authenticated_admin_user_gets_filtered_list_of_asset_logs_by_asset_make
         asset = Asset.objects.get(asset_code=asset_code)
         asset_make = asset.model_number.asset_make.name
         self.assertEqual(asset_make, self.asset_make.name)
+
+    @patch("api.authentication.auth.verify_id_token")
+    def test_authenticated_admin_user_asset_logs_by_return_detailed_fields(
```
kmazi: You can use a more descriptive name for this test method. Rather than writing out a new test, you could update an existing test for getting assetlogs to show that the categorization fields are included in the response.

kmazi: Will those fields be returned also when viewing a single log(detail)? It would be nice to show that in the test also

kmazi (changes requested):

api/serializers/assets.py
```diff
@@ -130,6 +140,39 @@ def to_internal_value(self, data):
             internals["specs"] = specs
         return internals
 
+    # # allow updating of active_inactive and paid for specific asset_types
+    def update(self, instance, validated_data):
+
+        asset_type = instance.model_number.asset_make.asset_type.name
+        paid = validated_data.get('paid')
+        active_inactive = validated_data.get('active')
```
raheemazeezabiodun, line 148: this seems confusing, can you give it a more declarative name, how about `is_active`?

What's the rationale behind the value of this variable?

kenneth051: @raheemazeezabiodun That is not part of this PR. I have a commit from develop that has mixed up with my commits...

api/serializers/assets.py
```diff
@@ -130,6 +140,39 @@ def to_internal_value(self, data):
...
+        expiry_date = validated_data.get('expiry_date')
+        instance_type_and_associated_error = {
+            "simcard": {"paid": "Only sim cards can have this field updated"},
+            "mifi": {"active": "Only mifi cards can be activated or deactivated"},
+            "embursecard": {
+                "expiry_date": "Only emburse cards can have this field updated"
+            },
+        }
+
+        if paid or expiry_date or (active_inactive is False or active_inactive is True):
```
raheemazeezabiodun, line 158: `active_inactive is False or active_inactive is True` this will always return True as long as `active_inactive` is a boolean value.

There is no need for `active_inactive is False or active_inactive is True`

Just do this to catch when active_inactive is None
```
if paid or expiry_date or active_inactive
```

kenneth051: @raheemazeezabiodun That is not part of this PR. I have a commit from develop that has mixed up with my commits...

api/serializers/assets.py
raheemazeezabiodun: cool cool

api/tests/test_asset_log.py
raheemazeezabiodun: cool cool

api/serializers/assets.py
raheemazeezabiodun: Alright

Although that part needs to be fixed, can you include it to your pr?

kenneth051: @raheemazeezabiodun No

api/tests/test_asset_log.py
kenneth051: @kmazi yes, they will be

kmazi (approved):

PR #310, opened 2019-06-28
remove security user email endpoint

kenneth051:
#### What does this PR do?
remove security user email endpoint

This PR removes the removes security user email endpoint

https://www.pivotaltracker.com/n/projects/2146417/stories/165931831


api/tests/test_security_user.py
```diff
@@ -29,35 +29,12 @@ def setUp(self):
 
         self.access_token = json.loads(response.content)["access_token"]
 
-    def test_app_can_get_security_users_emails(self):
+    def test_get_security_users_emails_endpoint_doesnt_exist(self):
```
kmazi: I don't think this test is necessary. Taking off the previous ones will do. Tests show the app's functionality. So when I don't see this endpoint being consumed in the tests, I know it's not available.

kmazi (changes requested):

api/tests/test_security_user.py
kenneth051: @kmazi Alright, let me remove it

kmazi (approved):

PR #315, opened 2019-07-05
Add date to Notifications endpoint

kenneth051:
#### What does this PR do?
This PR adds a `created_at` field in the incident model.

we are adding a `created_at` field to show when the incidents were created. we are also ordering the returned output using the created_at field. 

1. On postman, go to the  http://127.0.0.1:8000/api/v1/incidence-reports/. you will see the returned data has a `created_at` field.

ensure you apply the migrations by running the command, `python manage.py migrate`

https://www.pivotaltracker.com/n/projects/2146417/stories/160192851


api/tests/test_asset_incident_report_api.py
```diff
@@ -100,7 +100,37 @@ def test_authenticated_user_get_incident_report(self, mock_verify_id_token):
         self.assertEqual(
             len(response.data["results"]), AssetIncidentReport.objects.count()
         )
+        date = self.incident_report.created_at
```
kmazi: How many incident_reports are contained in the response such that this test shows the order by date(self.incident_report)?

kenneth051: There is only one incident report. Let me add another.

kmazi (approved):

PR #316, opened 2019-07-19
Elastic search assets and users 165965388

raheemazeezabiodun:
#### What does this PR do?
Allow full text search across different fields from the viewset

1. Create assets
2. search assets with this endpoint http://127.0.0.1:8000/api/v1/assets?search=whatever_you_want_to_search

[https://www.pivotaltracker.com/n/projects/2146417/stories/165965388](https://www.pivotaltracker.com/n/projects/2146417/stories/165965388)


api/tests/test_asset_api.py
```diff
@@ -387,3 +387,63 @@ def test_asset_allocation_history_has_assigner(self, mock_verify_id_token):
         )
         self.assertIn("assigner", response.data["allocation_history"][0].keys())
         self.assertEqual(response.status_code, 200)
+
+    @patch("api.authentication.auth.verify_id_token")
+    def test_asset_search_endpoint(self, mock_verify_id_token):
```
kmazi: Asset search is more of a functionality than an endpoint. However, there's an asset endpoint. A better description could be, "test_search_funx_on_asset_endpoint".

api/tests/test_asset_api.py
```diff
@@ -387,3 +387,63 @@ def test_asset_allocation_history_has_assigner(self, mock_verify_id_token):
...
+        # return only assets with ANDELA match which is just asset_one and asset_two
+        self.assertEqual(response.data["count"], 2)
+        self.assertEqual(
+            response.data.get('results')[1].get('uuid'), str(asset_one.uuid)
+        )
+        self.assertEqual(
+            response.data.get('results')[0].get('uuid'), str(asset_two.uuid)
+        )
+
+        response = client.get(
```
kmazi, line 524: It would be more appropriate if the tests are independent to show different functionalities rather than grouping them together. Different tests cases could should show different fields the search query string reaches.

kmazi: The test objects created could be done in a helper function/method and called when need be.

kmazi (changes requested):

PR #317, opened 2019-07-22
create history endpoint 

kenneth051:
#### What does this PR do?
Adds a history endpoint to the application

This endpoint enables an admin to track create, update and delete actions done in the application

1. Make a create, update or delete action and after navigate to the `http://127.0.0.1:8000/api/v1/history` history endpoint and make a get request. You will see all actions performed on the application.

after pulling the branch, Run migrations to have the history model in your database.

https://www.pivotaltracker.com/n/projects/2146417/stories/167008828


core/tests/test_history_model.py
```diff
@@ -0,0 +1,29 @@
...
+        History.objects.create(
+            table_name="core_asset",
+            user=self.user,
+            item_id="123",
+            action="POST",
+            body="body",
+        )
+        self.assertEqual(History.objects.count(), count + 1)
+
+    def test_cannot_add_history_with_nonexisting_user(self):
```
kmazi, line 34: This test was meant to show that you cannot add history with non-existing user. The description says the test is to assert that history cannot be created without action name. They could possibly be different tests you know.

core/tests/__init__.py
```diff
@@ -66,6 +66,14 @@ def setUpClass(cls):
             is_securityuser=True,
         )
 
+        cls.history = apps.get_model("core", "History").objects.create(
```
kmazi: It's obvious this would only be used in the HistoryModelTest test cases. It should be created and used there. Only objects that would be shared by other test cases should be created here. It can significantly improve the time it takes the tests to run.

core/models/history.py
```diff
@@ -0,0 +1,17 @@
+# Third-Party Imports
+from django.db import models
+
+
+class History(models.Model):
+    table_name = models.CharField(max_length=255, blank=False, editable=False)
+    user = models.ForeignKey("User", on_delete=models.PROTECT)
+    created_at = models.DateTimeField(auto_now_add=True, editable=False)
+    item_id = models.CharField(max_length=255, null=False, blank=False, editable=False)
+    action = models.CharField(max_length=25, null=False, blank=False, editable=False)
```
kmazi: If the action field represents "CREATE", "UPDATE" or "DELETE" then 25 as the max_length value is too high. I'm thinking it should be a choice field.

core/tests/__init__.py
```diff
@@ -66,6 +66,14 @@ def setUpClass(cls):
...
+            table_name="core_asset",
+            user=cls.user,
+            item_id="123",
+            action="POST",
+            body="body",
```
kmazi: What does the body field represent?

core/tests/test_history_model.py
kmazi: When testing for non-existing users, it's more appropriate to use a user object that isn't saved to the database for the test rather than a bare string.

api/tests/test_history_endpoint.py
```diff
@@ -0,0 +1,46 @@
...
+        response = client.get(
+            url, HTTP_AUTHORIZATION="Token {}".format(self.token_user)
+        )
+        self.assertEqual(response.data["id"], self.history1.id)
+        self.assertEqual(response.data["table_name"], self.history1.table_name)
+        self.assertEqual(response.data["item_id"], self.history1.item_id)
+        self.assertEqual(response.data["action"], self.history1.action)
+        self.assertEqual(response.data["user"], self.history1.user.id)
+        self.assertEqual(response.data["body"], self.history1.body)
+        self.assertEqual(response.status_code, 200)
```
kmazi, line 117: There should also be a test that shows that making post, update or delete requests from arbitrary endpoints create a history object.

api/tests/__init__.py
```diff
@@ -150,6 +150,13 @@ def setUpClass(cls):
             name="IMN50987345", asset_make=cls.asset_make
         )
 
+        cls.history1 = apps.get_model("core", "History").objects.create(
```
kmazi: It's more appropriate if this object is created in the `HistoryAPITest` test case since it's not shared by other test cases.

kmazi (changes requested):

api/signals.py
```diff
@@ -0,0 +1,27 @@
...
+from django.dispatch import receiver
+
+# App Imports
+from core.models import History
+
+
+@receiver(post_save)
+@receiver(post_delete)
+def track_application_actions(sender, instance, **kwargs):
+    current_request = CrequestMiddleware.get_request()
```
kmazi: Seems the only purpose of this package is to get the current request. We could easily create that with a middle-ware. Thus we won't be needing this package.

kmazi (changes requested):

api/signals.py
kenneth051: @kmazi Yeah, I have been working on this today, I will remove the package.

api/tests/__init__.py
kenneth051: yap. I get it . Thanks

core/tests/__init__.py
kenneth051: @kmazi the body field handles the request data being saved by the user.

core/tests/__init__.py
kenneth051: I have moved it

core/models/history.py
kenneth051: @kmazi, let me make it a choice field

api/tests/test_history_endpoint.py
kenneth051: done

core/tests/test_history_model.py
kenneth051: @kmazi  a bare string and none saved user object is still the same thing according to me

kmazi (approved):

PR #318, opened 2019-07-24
Get asset by user 167424021

raheemazeezabiodun:
#### What does this PR do?
It allows logged in user retrieve all asset owned

I should get a response which includes assets assigned to that particular user

1. On browser, goto <http://127.0.0.1:8000/api/v1/assets/me/>

Logged in user can be gotten from the token

https://www.pivotaltracker.com/n/projects/2146417/stories/167424021

### DIFF
--- a/api/tests/test_asset_api.py
+++ b/api/tests/test_asset_api.py
@@ -422,6 +422,17 @@ class AssetTestCase(APIBaseTestCase):
         )
 
     @patch("api.authentication.auth.verify_id_token")
+    def test_fetching_assets_assigned_to_non_existing_user(self, mock_verify_id_token):
+        """Admin should fetch asset detail for any given user."""
+        mock_verify_id_token.return_value = {"email": self.admin_user.email}
+        response = client.get(
+            f"{self.asset_urls}?user_id=1000",
+            HTTP_AUTHORIZATION="Token {}".format(self.token_admin),
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data.get("count"), 0)
+
+    @patch("api.authentication.auth.verify_id_token")
     def test_fetching_assets_assigned_to_a_random_user_by_non_admin(
         self, mock_verify_id_token
     ):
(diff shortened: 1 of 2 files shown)


core/models/__init__.py
```diff
@@ -8,6 +8,7 @@
     AssetLog,
     AssetMake,
     AssetModelNumber,
+    AssetOwner,
```
kmazi: Seems this is the same as `AssetAssignee` already imported on line 4.

api/views/assets.py
```diff
@@ -286,6 +286,19 @@ def get_queryset(self):
         return self.queryset.none()
 
 
+class UserAssetsView(APIView):
```
kmazi: I think creating a new view for getting user asset is unnecessary as it separates this functionality from the AssetViewset. Since we're getting assets in the response and the AssetViewset contains logic to create and return assets, it's more appropriate to have this logic there.

kmazi: We could leverage on marking [extra actions](https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing) in the AssetViewSet. We could have an extra action like `get-user-assets`

api/urls.py
```diff
@@ -115,6 +116,7 @@ def __init__(self, trailing_slash="/?"):
         name="api-version-index",
     ),
     path("export-assets/", ExportAssetsDetails.as_view(), name="export-assets"),
+    path("assets/me/", UserAssetsView.as_view(), name="assets-me"),
```
kmazi: This would be unnecessary if we leverage on extra actions in the AssetsViewSet

api/tests/test_asset_api.py
```diff
@@ -386,4 +387,35 @@ def test_asset_allocation_history_has_assigner(self, mock_verify_id_token):
...
+            asset_code="IC001453",
+            serial_number="SN5012345",
+            purchase_date="2018-07-10",
+            model_number=self.assetmodel,
+            asset_location=self.centre,
+            department=self.department,
+            verified=True,
+        )
+        response = client.get(
+            "{}".format(reverse("assets-me")),
```
kmazi: I'll look to restructure the tests. I feel it would be neater and more readable to use f-strings for our string concatenations.

kmazi (changes requested):

PR #320, opened 2019-07-30
ft-add-department-teams: add departmental teams

gitau-kinuthia:
#### What does this PR do?
Add functionality for adding departmental teams and assigning assets to them

- add teams model
- allow users to be  assigned to teams
- add functionality to allow assignments of assets to teams
- add functionality to view assets assigned to a team

1. On browser, goto url `http://127.0.0.1:8000/api/v1/docs/live/` to access all endpoints
2. Scroll to the `department-teams` section to view all associated endpoints for CRUD
`http://127.0.0.1:8000/api/v1/department-teams/`
`http://127.0.0.1:8000/api/v1/department-teams/<id>`

[#167427843](https://www.pivotaltracker.com/story/show/167427843)

image

image

image


api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,101 @@
...
+    @patch("api.authentication.auth.verify_id_token")
+    def test_authenticated_user_get_department_teams(self, mock_verify_token):
+        mock_verify_token.return_value = {"email": self.admin_user.email}
+        response = client.get(
+            self.departmental_teams_url,
+            HTTP_AUTHORIZATION="Token {}".format(self.token_user),
+        )
+        data = response.json()
+        assert isinstance(data['results'], list)
+        assert response.status_code == status.HTTP_200_OK
```
kmazi: It's ok to let at least one of your test method show all the dictionary key and value expected from hitting the `departmental_teams_url` endpoin in the assertions.. Then other ones could show only specific assertions particular to the test method.

api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @patch("api.authentication.auth.verify_id_token")
+    def test_authenticated_user_get_specific_department_teams_no_data(
```
kmazi: What does `no_data` in the test method name mean?

api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,101 @@
...
+        self, mock_verify_token
+    ):
+        mock_verify_token.return_value = {"email": self.admin_user.email}
+        url = "{}/{}/".format(self.departmental_teams_url, self.departmental_team_1.id)
+        response = client.get(
+            url, HTTP_AUTHORIZATION="Token {}".format(self.token_user)
+        )
+        data = response.json()
+        assert response.status_code == status.HTTP_200_OK
+        assert "assets_assigned" in data
```
kmazi: Your assertions could show the expected fields when querying for a specific department team just like it should when querying for all teams of a department.

api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,101 @@
...
+        allocation = AllocationHistory.objects.create(
+            asset=new_asset, current_assignee=self.asset_assignee_team
+        )
+        # authenticate a user
+        mock_verify_token.return_value = {"email": self.admin_user.email}
+        url = "{}/{}/".format(self.departmental_teams_url, self.departmental_team.id)
+        team_response = client.get(
+            url, HTTP_AUTHORIZATION="Token {}".format(self.token_user)
+        )
+        asset = AssetSerializer(new_asset).data
```
kmazi: This is unnecessary I suppose; We could do without it.

api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,101 @@
...
+            "description": "string",
+            "department": self.department.id,
+        }
+        response = client.post(
+            self.departmental_teams_url,
+            data=payload,
+            HTTP_AUTHORIZATION="Token {}".format(self.token_user),
+        )
+        data = response.json()
+        assert isinstance(data, dict)
```
kmazi: What keys are returned after a post request? You test could show that.

settings/base.py
```diff
@@ -33,16 +33,15 @@
...
 # See https://docs.djangoproject.com/en/2.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
 
 SECRET_KEY = config("SECRET_KEY")
 
-DATABASES = {"default": dj_database_url.config()}
-
+# DATABASES = {"default": dj_database_url.config()}
+DATABASES = {"default": dj_database_url.config(default=config('DATABASE_URL'))}
```
kmazi: Why is this necessary?

core/models/centre.py
```diff
@@ -167,3 +167,15 @@ class Meta:
 
     def __str__(self):
         return self.name
+
+
+class DepartmentalTeam(models.Model):
```
kmazi, line 172: There seem to be no tests for the expected behavior of the `DepartmentalTeam` model.

api/tests/__init__.py
```diff
@@ -240,6 +249,7 @@ def setUpClass(cls):
         cls.centre_url = reverse("andela-centres-list")
         cls.country_url = reverse("countries-list")
         cls.department_url = reverse("departments-list")
+        cls.departmental_teams_url = reverse("departmental-teams-list")
```
kmazi: This is used only in the DepartmentTeamTestCase. We could move it there. Cleaning up this setup class can help improve the run time of tests significantly.

api/tests/__init__.py
```diff
@@ -171,6 +177,9 @@ def setUpClass(cls):
         cls.asset_assignee = apps.get_model("core", "AssetAssignee").objects.get(
             user=cls.user
         )
+        cls.asset_assignee_team = apps.get_model(
```
kmazi: This can move also to the DepartmentTeamTestCase `setUp` method.

api/tests/__init__.py
```diff
@@ -59,6 +59,12 @@ def setUpClass(cls):
         cls.department_travel = apps.get_model("core", "Department").objects.create(
             name="Travel"
         )
+        cls.departmental_team = apps.get_model(
```
kmazi: Both can move to the DepartmentTeamTestCase setUp method.

api/serializers/andela_centres.py
```diff
@@ -196,3 +199,56 @@ class CountrySerializer(serializers.ModelSerializer):
...
+    def get_assets_assigned(self, obj):
+        """
+        list all assets assigned to  a specific team
+        :param obj:
+        :return:
+        """
+
+        team_assignee = models.AssetAssignee.objects.filter(team=obj).first()
+        # The following condition introduced to make sure that results are not returned if there are orphan
+        # records in the that have no assignee (assigned_to field is optional so its possible to have an empty
```
kmazi: Good doc though with slight typo error.

api/serializers/andela_centres.py
```diff
@@ -196,3 +199,56 @@ class CountrySerializer(serializers.ModelSerializer):
...
+        if team_assignee:
+            assets = models.Asset.objects.filter(assigned_to=team_assignee)
+            page = self.context["view"].paginate_queryset(assets)
+            serialized_assets = DepartmentAssetSerializer(page, many=True)
+            paginated_assets = self.context["view"].get_paginated_response(
+                serialized_assets.data
+            )
+            return paginated_assets.data
+        # return a standardised response to the api with the same structure as when there are results
+        return {"count": 0, "next": None, "previous": None, "results": []}
```
kmazi, line 239: Nice work.

.circleci/config.yml
```diff
@@ -98,6 +98,11 @@ jobs:
       - *cmd_restore_cache
       - attach_workspace:
           at: ~/art-backend/tmp
+      - run:
+          name: run migrations
+          command: |
+            pipenv run python manage.py makemigrations
```
kmazi: Why make migrations on circle ci again? All you need to do is run migration that has been made locally.

kmazi (changes requested):
You could also show in your test that when querying the user endpoint, the response also contains the team the user belongs to.

.circleci/config.yml
gitau-kinuthia: Absence of this was the error that was causing my tests to fail on Circle CI. Once I added this line of code my tests could pass without throwing psycopg2 errors.

Reference : https://stackoverflow.com/questions/48184300/when-in-run-test-cases-then-i-will-get-this-error-psycopg2-operationalerror-cu?rq=1

api/tests/test_departmental_teams_api.py
gitau-kinuthia: The test has been renamed to `test_authenticated_user_get_specific_department_teams_with_no_assets_assigned` to be more communicative

core/models/centre.py
gitau-kinuthia: This have been added

api/tests/test_departmental_teams_api.py
gitau-kinuthia: This test case is necessary to show the team_name field exists in the assets model. I have refactored this to a separate test to ensure test autonomy

kmazi: Ok. I was referring to calling the serializer here. We could just read off the new_asset data directly for assertion.

api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,171 @@
...
+            HTTP_AUTHORIZATION="Token {}".format(self.token_user),
+        )
+        data = response.json()
+        # return types assertions
+        assert isinstance(data['results'], list)
+        assert isinstance(data['results'][0], dict)
+        # results field assertions
+        assert 'name' in data['results'][0]
+        assert 'description' in data['results'][0]
+        assert 'department' in data['results'][0]
```
kmazi: A better assertion would be `self.assertEqual(data["results"][0]["department"], self.department_team.department)`. Let's assert for the values of those keys rather than the keys. We can do this for the other tests.

api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,171 @@
...
+    @patch("api.authentication.auth.verify_id_token")
+    def test_authenticated_user_get_department_teams(self, mock_verify_token):
+        mock_verify_token.return_value = {"email": self.admin_user.email}
+        response = client.get(
+            self.departmental_teams_url,
+            HTTP_AUTHORIZATION="Token {}".format(self.token_user),
+        )
+        data = response.json()
+        # return types assertions
+        assert isinstance(data['results'], list)
```
kmazi: This is pytest assertion style. Would this run if the test is run with the `python manage.py test` command?

kmazi: Let's make the changes so that both `pytest` command and `python manage.py test` command works for all tests.

api/tests/test_departmental_teams_api.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+        mock_verify_token.return_value = {"email": self.admin_user.email}
+        url = "{}/{}/".format(self.departmental_teams_url, self.departmental_team.id)
+        team_response = client.get(
+            url, HTTP_AUTHORIZATION="Token {}".format(self.token_user)
+        )
+        data = team_response.json()
+        assert "members" in data
+        assert len(data["members"]["results"]) == 1
+        assert data["members"]["results"][0]["email"] == new_user.email
```
kmazi: Great!! These assertions are super cool.

core/tests/test_department_team.py
```diff
@@ -0,0 +1,40 @@
...
+            name="Accounts", department_id=self.new_department.id
+        )
+        new_team.save()
+        assert DepartmentalTeam.objects.count() == count + 1
+        new_team.delete()
+
+    def test_cannot_add_existing_team(self):
+        team = DepartmentalTeam(name="Accounts", department_id=self.new_department.id)
+        team.save()
+        with transaction.atomic():
```
kmazi: The default TestCase wraps all tests methods in a transaction. There is no real need to explicitly manage the transactions here in the tests.

core/tests/test_department_team.py
```diff
@@ -0,0 +1,40 @@
...
+            with pytest.raises(utils.IntegrityError):
```
kmazi: This also is germane to pytest. Would this work with `python manage.py test`? Unless we're planning to deprecate running with django test.

kmazi (changes requested):

core/tests/test_department_team.py
gitau-kinuthia: @kmazi the atomic block is actually necessary. In its absence, the following error is obtained:
```django.db.transaction.TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.```

core/tests/test_department_team.py
gitau-kinuthia: Yes the tests still run using the ```python manage.py test```

api/tests/test_departmental_teams_api.py
gitau-kinuthia: > This is pytest assertion style. Would this run if the test is run with the `python manage.py test` command?

Yes

core/tests/test_department_team.py
kmazi: My point is that we don't need a db transaction for this test method especially because the TestCase base class handles transaction for all test methods. Therefore calling it explicitly is unnecessary. @fgkinus

core/tests/test_department_team.py
gitau-kinuthia: This has been refactored however to the unittest assertion  style

api/tests/test_departmental_teams_api.py
gitau-kinuthia: > Let's make the changes so that both `pytest` command and `python manage.py test` command works for all tests.

This has been refactored however to the unittest assertion  style

api/tests/test_departmental_teams_api.py
gitau-kinuthia: This has been refactored however to the unittest assertion  style

core/tests/test_department_team.py
gitau-kinuthia: Noted. I have been able to remove the delete function calls that were eliciting the transaction errors.

kmazi (approved):

PR #327, opened 2019-08-22
#161678177 : Add notifications model

gitau-kinuthia:
#### What does this PR do?
- Add Notifications Model
- configure django_q and django_cache

- Add a notification model based on the nomenclature [here](https://www.w3.org/TR/notifications/)
- Add signals to create notifications to admins when a new incident report is filed
- Add signals to notify the user when the transition state for an incident report raised is changed

- To setup the Django cache which is required for  monitoring `django_q` ([details](https://docs.djangoproject.com/en/1.8/topics/cache/#database-caching)) run: 
```
pipenv run python manage.py createcachetable
```

- The nomenclature for the notifications is based on the spec in www.w3.org/TR/notifications/. 
The expected body of notifications is :
Generalized Format for notifications body: as proposed in the [django_notifications ](https://github.com/django-notifications/django-notifications/blob/master/notifications/base/models.py)package
 ```
        <actor> <verb> <time>

        <actor> <verb> <target> <time>

        <actor> <verb> <action_object> <target> <time>
```

[161678177](https://www.pivotaltracker.com/story/show/161678177)

### CODE REVIEW

File: api/signals.py [hunk 1]
```diff
@@ -30,3 +42,61 @@ def track_application_actions(sender, instance, **kwargs):
...
+def create_notification_on_incident_report_submission(**kwargs):
+    """
+    Create a notification to admin users when a user creates an incident report
+    :param kwargs:
+    :return:
+    """
+    instance = kwargs["instance"]
+    # if a new notification has been  created
+    if kwargs.get("created", False) is True:
+        notification_body = "{} submitted a new incident report {}."
```
kenneth051: @fgkinus good work, I have suggestions. the notification details could be added in the constant.py file. we would just call them from there then format them if we need to add contents to them. I mean the notification body, title. we could also do away with line 60 and just add the instance.submitted_by directly to origin on line 69.

api/signals.py
```diff
@@ -30,3 +42,61 @@ def track_application_actions(sender, instance, **kwargs):
...
+@receiver(post_save, sender=StateTransition)
+def alert_user_when_incident_report_status_is_updated(**kwargs):
+    """
+    When the status of the incident report is modified, alert the user that filed the report
+    :param kwargs:
+    :return:
+    """
+    instance = kwargs["instance"]
+    # notification details
+    incident_url = "{}/{}".format(reverse("incidence-reports-list"), instance.id)
```
kenneth051: we could also add these to the constant.py file

core/models/notification.py
```diff
@@ -0,0 +1,84 @@
...
+        <actor> <verb> <action_object> <target> <time>
+
+    Examples::
+        <user> <reached level 60> <1 minute ago>
+        <user2> <commented on> <url/post/1d> <2 hours ago>
+        <user> <started follow> <user2> <8 minutes ago>
+        <user> <closed> <issue 70> on <topic> <about 2 hours ago>
+    """
+
+    title = models.CharField(
```
kenneth051, line 28: I also see that the model fields specified in the story are not the same as you have here.. could you reach out to the FPC and have the story description updated.

core/tests/test_asset_incident_report_model.py
```diff
@@ -23,3 +28,48 @@ def test_model_string_representation(self):
...
+            self.admin_user.email, incident_url
+        )
+
+        self.assertEqual(
+            admin_count, notifications.count() - original_notifications_count
+        )
+        self.assertEqual(notification.origin, report.submitted_by)
+        self.assertEqual(notification.title, "New Incident Report")
+
+        self.assertEqual(notification.body, body)
```
kenneth051, line 74: Is there a way to test for failure when a notification is not sent.

api/signals.py
gitau-kinuthia: Thanks for these notes. I agree and  will incorporate the feedback 👍

core/models/notification.py
gitau-kinuthia: I will reach out and have the story updated 👍

core/tests/test_asset_incident_report_model.py
gitau-kinuthia: I think to test when a notification not sent would entail counting notifications before triggering an event that would usually generate a notification and then verifying that notification count does not change.

core/tests/test_asset_incident_report_model.py
```diff
@@ -23,3 +28,48 @@ def test_model_string_representation(self):
             str(incident_report),
             f"{incident_report.incident_type}: {incident_report.asset}",
         )
+
+
+class TestAssetIncidentReportNotifications(CoreBaseTestCase):
```
kmazi, line 33: Nice categorization of test cases.

core/tests/test_asset_incident_report_model.py
```diff
@@ -23,3 +28,48 @@ def test_model_string_representation(self):
...
+    """
+    Test Notifications elicited by Asset Incident Report creation
```
kmazi: Just for semantics sake and nothing related to tech, `elicited` isn't the right word to use there. looool

core/tests/test_asset_incident_report_model.py
```diff
@@ -23,3 +28,48 @@ def test_model_string_representation(self):
...
+        original_notifications_count = self.Notifications_model.objects.count()
+        admins = self.User_model.objects.filter(is_superuser=True)
+        admin_count = admins.count()
+        report = AssetIncidentReport.objects.create(
+            asset=self.test_asset, incident_type=constants.LOSS, submitted_by=self.user
+        )
+        notifications = self.Notifications_model.objects.filter(target__in=admins)
+        notification = notifications[0]
+        incident_url = "{}/{}".format(reverse("incidence-reports-list"), report.id)
+        body = "{} submitted a new incident report {}.".format(
```
kmazi: Let's stick to [fstrings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) for formatting. @kenneth051 what's your take on this?

core/models/notification.py
```diff
@@ -0,0 +1,84 @@
...
+from django.db import models
+from django.utils.timesince import timesince
+
+# App Imports
+from core import constants
+
+
+class Notifications(models.Model):
+    """
+    A model representing the a notification object
```
kmazi: slight typo

core/models/notification.py
kmazi: Let's tailor the generalized format to our specific use case as much as possible. Nice job

core/tests/test_asset_incident_report_model.py
kmazi: Appears to me like this test would serve that purpose since it fails when notification is not sent. @kenneth051

core/tests/test_notifications_model.py
```diff
@@ -0,0 +1,32 @@
...
+        self.assertEqual(self.notification.read, True)
+
+    def test_mark_notification_as_unread(self):
+        self.notification.mark_as_read()
+        self.notification.mark_as_unread()
+        self.assertEqual(self.notification.read, False)
+
+    def test_get_time_since(self):
+        timesince = self.notification.timesince()
+        self.assertTrue(isinstance(timesince, str))
```
kmazi, line 32: This assertion doesn't show the time since notification was created but rather asserts the type.

core/tests/test_asset_incident_report_model.py
gitau-kinuthia: Noted and fixed

core/constants.py
```diff
@@ -121,3 +120,14 @@
 SIMCARD_ASSET_OPTIONS = ((BASIC, "basic"), (BUSINESS, "business"), (NONE, "none"))
 
 TABLES = ["core_history", "django_admin_log"]
+
+NOTIFICATION_TITLES = ()
```
kmazi: What's the essence of this constant?

api/signals.py
```diff
@@ -30,3 +42,61 @@ def track_application_actions(sender, instance, **kwargs):
...
+    # if a new notification has been  created
+    if kwargs.get("created", False) is True:
+        notification_body = "{} submitted a new incident report {}."
+        notification_title = "New Incident Report"
+        notification_targets = User.objects.filter(is_superuser=True)
+        notification_origin = instance.submitted_by
+        incident_url = "{}/{}".format(reverse("incidence-reports-list"), instance.id)
+
+        # create notifications to te different admins
+        for target in notification_targets:
```
kmazi: This is a good place to use [atomic commit](https://docs.djangoproject.com/en/2.2/topics/db/transactions/#django.db.transaction.atomic) to ensure this is done with a single call to the database in a transaction.

api/signals.py
```diff
@@ -30,3 +42,61 @@ def track_application_actions(sender, instance, **kwargs):
...
+def alert_user_when_incident_report_status_is_updated(**kwargs):
+    """
+    When the status of the incident report is modified, alert the user that filed the report
+    :param kwargs:
+    :return:
+    """
+    instance = kwargs["instance"]
+    # notification details
+    incident_url = "{}/{}".format(reverse("incidence-reports-list"), instance.id)
+    notification_body = "The status of incident report {} that you filed has been updated to `{}`.".format(
```
kmazi: `F-strings` is kinda cleaner in string literals interpolation.

kmazi (changes requested):

core/constants.py
gitau-kinuthia: It was supposed to be a tuple containing valid `Notification Titles` for the notifications. title field. I decided against this approach for flexibility sake but didn't remove the variable. I have fixed this though

api/signals.py
gitau-kinuthia: Thanks for this feedback. This actually something I had not considered  👍

core/tests/test_notifications_model.py
gitau-kinuthia: Any ideas on a test that would work. Knowing that the current time will be an unknown variable i.e I cannot think of a way to capture the actual time the test runs so as to do a value assertion

gitau-kinuthia: > This assertion doesn't show the time since notification was created but rather asserts the type.
I figured a type assertion would make sure would make sure that a string representaion of time is actually returned veryfing the function works

core/constants.py
```diff
@@ -121,7 +122,16 @@
 
 TABLES = ["core_history", "django_admin_log"]
 
-NOTIFICATION_TITLES = ()
+# Notification titles
+INCIDENT_REPORT_CREATED_NOTIFICATION_TITLE = "New Incident Report"
+INCIDENT_REPORT_STATUS_UPDATED_NOTIFICATION_TITLE = "Incident Report Status Update"
+# Notification Bodies
+INCIDENT_REPORT_CREATED_NOTIFICATION_BODY = Template(
+    "$email submitted a new incident report {$link}."
```
kmazi: Are the curly braces intentional seeing that they would display in the resulting string?

kmazi (approved):

core/constants.py
gitau-kinuthia: @kmazi Thanks for the noting the mistake. I have removed them

kenneth051 (approved):

PR #329, opened 2019-09-01
Ft admin asset Type threshold notification

kenneth051:
[Finishes #167967311]

#### What does this PR do?
alerts admins in case the asset type threshold drops below the default threshold

Current, we are assigning assets to users and departments without keeping track of the current remaining assets. this PR addresses this issue and if the remaining assets go below the default asset threshold, then admins will be notified via email

1. on postman, visit the endpoint ` http://127.0.0.1:8000/api/v1/allocations`then assign an asset to a user

start by migrating the models. run the command python manage.py migrate to apply the necessary changes.
Make sure in ur `.env` file you set the variable `DEFAULT_THRESHOLD`to a number for instance 20. then in the `assettype` model edit the fields by giving them a threshold value that we will compare with the DEFAULT_THRESHOLD.
set the email fields in your .env and give them values
MAIL_HOST = 'smtp.gmail.com'
EMAIL_HOST_USER = " "
EMAIL_HOST_PASSWORD = " "
EMAIL_PORT = 587
EMAIL_USE_TLS = True
EMAIL_SENDER=" "
after starting the server, make sure you open another terminal tab and run the command `python manage.py qcluster`  to start the `qCluster`

https://www.pivotaltracker.com/n/projects/2146417/stories/167967311


core/models/asset.py
```diff
@@ -542,6 +545,13 @@ def save(self, *args, **kwargs):
             self.previous_assignee = None
         try:
             super().save(*args, **kwargs)
+            if self.previous_assignee is None:
+                threshold_data = self.asset.model_number.asset_make.asset_type
```
kmazi, line 549: This variable would be more descriptive if named `asset_type`. Then you can have asset_type.threshold = ....

api/send_email.py
```diff
@@ -0,0 +1,19 @@
...
+# App Imports
+from core.constants import MSG, SUBJECT
+from core.models.user import User
+
+
+def send_email(data):
+    """"This method sends emails to  users"""
+    for user in User.objects.filter(is_staff=True):
+        async_task(
+            "django.core.mail.send_mail",
```
kmazi, line 14: Consider using [send_messages](https://docs.djangoproject.com/en/2.2/topics/email/#sending-multiple-emails) for sending mails. It's more efficient.

kmazi (changes requested):

